PR #53, opened 2015-01-12
Crammd5

jdef:
Ported the "authenticatee" implementation from mesos native. There's a pretty clear path forward here for integrating support for other mechanisms. Also started integrating support for [Context][1], with the intent to support timeouts at various layers (including authentication) at some point in the future.

TODO:
- [x] should probably `Stop()` the authentication messenger at some point
- [x] authentication should abort if context is cancelled
- [x] abstract authentication mechanism (cc @vladimirvivien)
- [x] passing all unit tests
- [ ] additional unit tests for new authentication packages (WIP)
- [x] document StepFunc, as per https://github.com/mesos/mesos-go/pull/53#discussion_r22841074
- [x] rename: `authenticateeProcess.result` --> `authenticateeProcess.err`

[1]: https://blog.golang.org/context


auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	if err := self.transport.Send(ctx, from, message); err != nil {
+		self.terminate(status, ERROR, err)
+	}
+}
+
+func (self *authenticateeProcess) completed(ctx context.Context, from *upid.UPID, pbMsg proto.Message) {
+	status := statusFrom(ctx)
+	if status != STEPPING {
+		self.terminate(status, ERROR, fmt.Errorf("Unexpected authentication 'completed' received"))
```
jdef: extract constant (error)

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,320 @@
...
+		return
+	}
+
+	log.Infof("Authentication success")
```
jdef: s/Infof/Info

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func (self *authenticateeProcess) step(ctx context.Context, from *upid.UPID, pbMsg proto.Message) {
+	status := statusFrom(ctx)
+	if status != STEPPING {
+		self.terminate(status, ERROR, UnexpectedAuthenticationStep)
+		return
+	}
+
+	log.Infof("Received SASL authentication step")
```
jdef: s/Infof/Info

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// returns true when handlers are installed without error, otherwise terminates the
+// authentication process.
+func (self *authenticateeProcess) installHandlers(ctx context.Context) error {
+
+	type handlerFn func(ctx context.Context, from *upid.UPID, pbMsg proto.Message)
+
+	withContext := func(f handlerFn) messenger.MessageHandler {
+		return func(from *upid.UPID, m proto.Message) {
```
jdef: would be nice to get rid of another TODO, check `from` against an optional `fromFilter` here, to ensure that we keep talking to the same authenticator

auth/interface.go
```diff
@@ -0,0 +1,34 @@
...
+
+type AuthenticateeFunc func(ctx context.Context, pid, client upid.UPID, credendial mesos.Credential) <-chan error
+
+func (f AuthenticateeFunc) Authenticate(ctx context.Context, pid, client upid.UPID, creds mesos.Credential) <-chan error {
+	return f(ctx, pid, client, creds)
+}
+
+var (
+	AuthenticationFailed = errors.New("authentication failed")
+	UnsupportedMechanism = errors.New("failed to identify a compatible mechanism")
```
jdef: move this into the sasl package

auth/crammd5/authenticatee.go
```diff
@@ -0,0 +1,236 @@
...
+		client:         client,
+		credential:     credential,
+		status:         READY,
+		done:           make(chan struct{}),
+		ctx:            ctx,
+		cancel:         cancel,
+	}
+}
+
+func (self *authenticateeProcess) _fail(s statusType, err error) {
```
vladimirvivien: Maybe call this method just fail() instead of _fail()

auth/crammd5/authenticatee.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func nextPid() int {
+	pidLock.Lock()
+	defer pidLock.Unlock()
+	pid++
+	return pid
+}
+
+func Authenticatee(pid, client upid.UPID, credential mesos.Credential) <-chan error {
```
vladimirvivien: Maybe rename NewAuthenticatee()  to indicate its a constructor function.?

auth/crammd5/authenticatee.go
jdef: Agree: `_fail` was terrible. I ended up renaming it `terminate`

auth/crammd5/authenticatee.go
jdef: It's not a constructor function - just a top-level func in the package to handle the client side of the SASL authentication process. For comparison `newAuthenticatee` is a constructor function.

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,320 @@
...
+	pid     int
+
+	UnexpectedAuthenticationMechanisms = errors.New("Unexpected authentication 'mechanisms' received")
+	UnexpectedAuthenticationStep       = errors.New("Unexpected authentication 'step' received")
+)
+
+type statusType int32
+
+const (
+	READY statusType = iota
```
jdef: there's no sense exporting these since `statusType` is package private. consider renaming, for example, `READY` to something like `statusReady`

auth/sasl/context.go
```diff
@@ -0,0 +1,33 @@
...
+type _key int
+
+// If this package defined other context keys, they would have
+// different integer values.
+const (
+	statusKey _key = iota
+)
+
+var (
+	NilProcessErr = errors.New("nil authenticateeProcess has no status")
```
jdef: remove unused error code

auth/sasl/mech/crammd5/mechanism.go
```diff
@@ -0,0 +1,67 @@
...
+	"errors"
+	"io"
+
+	log "github.com/golang/glog"
+	"github.com/mesos/mesos-go/auth/sasl/callback"
+	"github.com/mesos/mesos-go/auth/sasl/mech"
+)
+
+var (
+	illegalStateErr = errors.New("illegal mechanism state")
```
jdef: Should this be moved to the `mech` package and exported (along with `illegalState` step function)?

auth/sasl/mech/plugins.go
```diff
@@ -0,0 +1,39 @@
...
+
+	if _, found := supportedMechs[name]; found {
+		return fmt.Errorf("Mechanism registered twice: %s", name)
+	}
+	supportedMechs[name] = f
+	log.V(1).Infof("Registered mechanism %s", name)
+	return nil
+}
+
+func SelectSupported(mechanisms []string) (selectedMech string, factory Factory) {
```
jdef: Ideally this would return a set of all matching names, and the client would then decide which one to use, then fetch it using a (currently non-existent) call like `GetFactory(selectedMech) (Factory, bool)`

auth/sasl/mesos/handlers.go
```diff
@@ -0,0 +1,24 @@
+package mesos
+
+import (
+	"github.com/mesos/mesos-go/auth/sasl/callback"
+	mesos "github.com/mesos/mesos-go/mesosproto"
+)
+
+type CredentialHandler struct {
```
jdef: rename this to `UsernamePasswordHandler`, or something along those lines

auth/crammd5/authenticatee.go
```diff
@@ -0,0 +1,236 @@
...
+	done       chan struct{}
+	result     error
+	ctx        context.Context
+	cancel     context.CancelFunc
+	mech       *mechanism
+	stepFn     stepFunc
+}
+
+func init() {
+	supportedMechs = make(map[string]struct{})
```
vladimirvivien: The map[string]struct{} wasn't clear until I saw the search loop in mechanism() method.

auth/crammd5/mechanism.go
```diff
@@ -0,0 +1,42 @@
...
+		return illegalState, nil, err
+	}
+
+	codes := hex.EncodeToString(hash.Sum(nil))
+	msg := m.username + " " + codes
+	return nil, []byte(msg), nil
+}
+
+func illegalState(m *mechanism, data []byte) (stepFunc, []byte, error) {
+	return illegalState, nil, illegalStateErr
```
vladimirvivien: Shoud illegalState() return nil, nil, illegalStateErr to avoid calling itself always ?

auth/crammd5/authenticatee.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+type authenticateeProcess struct {
+	*messenger.MesosMessenger
+	credential mesos.Credential
+	client     upid.UPID
+	status     statusType
+	initOnce   sync.Once
+	done       chan struct{}
+	result     error
```
vladimirvivien: Change the name of 'result' to say 'err' ? Since result will always return an error type.

auth/crammd5/authenticatee.go
jdef: `map[string]struct{}` is a simple way to implement a "set" of strings in go. The newer code maps `string` to `mech.Factory` so it's a bit more intuitive.

auth/crammd5/mechanism.go
jdef: Good question. I debated this for a bit myself, then failed to document my thinking: the intent is that `StepFunc` implementations should never return a `nil` StepFunc result. This makes the logic in the SASL authticatee simpler: step functions are never `nil`. Step functions that end up an error state (for example, some decoding logic fails...) should return a StepFunc that represents an error state. Some mechanisms may be able to recover from such. Not CRAMMD5. I'll add a TODO for docs around StepFunc.

auth/crammd5/authenticatee.go
jdef: Sure, that's more idiomatic.

auth/sasl/mesos/handlers.go
jdef: changed my mind, leaving as is

auth/sasl/mech/plugins.go
jdef: changed my mind, leaving as is

jdef:
Commit e34b1a0 made the client interface a bit more generic and added support for pluggable authtication providers (previously only SASL mechanisms were pluggable), which are also configurable via the command line. Hopefully this establishes a clear path forward for those (including us) implementing additional authentication providers down the road.

messenger/messenger.go
```diff
@@ -189,30 +194,71 @@ func (m *MesosMessenger) encodeLoop() {
...
 			return
 		case msg := <-m.sendingQueue:
-			if err := m.tr.Send(msg); err != nil {
-				errMsg := fmt.Sprintf("Failed to send message %v: %v", msg.Name, err)
-				log.Errorf(errMsg)
-				m.Route(m.UPID(), &mesos.FrameworkErrorMessage{Message: proto.String(errMsg)})
-				continue
+			e := func() error {
+				//TODO(jdef) implement timeout for context
+				ctx, cancel := context.WithCancel(context.Background())
```
jdef: use context.TODO() instead of Background()?

executor/executor.go
```diff
@@ -221,6 +223,23 @@ func (driver *MesosExecutorDriver) reregistered(from *upid.UPID, pbMsg proto.Mes
 	driver.exec.Reregistered(driver, slaveInfo)
 }
 
+func (driver *MesosExecutorDriver) send(upid *upid.UPID, msg proto.Message) error {
+	//TODO(jdef) should implement timeout here
+	ctx, cancel := context.WithCancel(context.Background())
```
jdef: use context.TODO instead?

messenger/messenger.go
```diff
@@ -189,30 +194,71 @@ func (m *MesosMessenger) encodeLoop() {
...
 		case msg := <-m.encodingQueue:
-			b, err := proto.Marshal(msg.ProtoMessage)
-			if err != nil {
-				errMsg := fmt.Sprintf("Failed to send message %v: %v", msg, err)
-				log.Errorln(errMsg)
-				m.Route(m.UPID(), &mesos.FrameworkErrorMessage{Message: proto.String(errMsg)})
-				continue
+			e := func() error {
+				//TODO(jdef) implement timeout for context
+				ctx, cancel := context.WithCancel(context.Background())
```
jdef: use context.TODO?

messenger/messenger.go
```diff
@@ -189,30 +194,71 @@ func (m *MesosMessenger) encodeLoop() {
...
-			msg.Bytes = b
-			m.sendingQueue <- msg
 		}
 	}
 }
 
+func (m *MesosMessenger) reportError(err error) {
+	log.V(2).Info(err)
+	//TODO(jdef) implement timeout for context
+	ctx, cancel := context.WithCancel(context.Background())
```
jdef: use context.TODO?

scheduler/scheduler.go
```diff
@@ -295,6 +317,23 @@ func (driver *MesosSchedulerDriver) resourceOfferRescinded(from *upid.UPID, pbMs
 	driver.Scheduler.OfferRescinded(driver, msg.OfferId)
 }
 
+func (driver *MesosSchedulerDriver) send(upid *upid.UPID, msg proto.Message) error {
+	//TODO(jdef) should implement timeout here
+	ctx, cancel := context.WithCancel(context.Background())
```
jdef: use context.TODO

scheduler/scheduler.go
```diff
@@ -151,6 +169,10 @@ func (driver *MesosSchedulerDriver) init() error {
 	driver.messenger.Install(driver.slaveLost, &mesos.LostSlaveMessage{})
 	driver.messenger.Install(driver.frameworkMessageRcvd, &mesos.ExecutorToFrameworkMessage{})
 	driver.messenger.Install(driver.frameworkErrorRcvd, &mesos.FrameworkErrorMessage{})
+
+	// install default authentication providers; extensions of this scheduler implementation
+	// are responsible for registering any additional providers they want to support.
+	auth.RegisterAuthenticateeProvider(saslProvider, auth.AuthenticateeFunc(sasl.Authenticatee))
```
jdef: move this into auth/sasl/plugin.go::init(), and then just import the auth/sasl package anonymously here so that it registers itself

auth/login.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// unexported key type, avoids conflicts with other context-using packages
+type loginKeyType int
+
+const (
+	loginNameKey loginKeyType = iota // name of login provider to use, e.g. SASL
+)
+
+func WithLoginName(ctx context.Context, name string) context.Context {
```
vladimirvivien: Short function description needed.

auth/login.go
```diff
@@ -0,0 +1,55 @@
...
+	ch <- err
+	close(ch)
+	return ch
+}
+
+// unexported key type, avoids conflicts with other context-using packages
+type loginKeyType int
+
+const (
+	loginNameKey loginKeyType = iota // name of login provider to use, e.g. SASL
```
vladimirvivien: To clear up the purpose of loginNameKey, I would change the identifier name to SaslLoginKey

jdef: loginName ~= loginProviderName. documentation definitely needed. Has nothing to do with SASL. The login provider thing was meant to abstract away the SASL/mechanism based interfaces. It's a small abstraction, but possibly useful. We can ditch it if you think it's going overboard. Are you familiar with Java's JAAS at all? The Login interface is somewhat inspired by JAAS LoginContext.

What's really ugly to me is how I return a `<-chan error` instead of just an `error` for Login and Authenticatee. Initially I was trying to emulate the "promise" API from the mesos implementation but I don't think it has a place in this Go-based API. If someone wants Login to run concurrently then they can just `go func() { ch <- auth.Login(...) }()` and use their own channel to process the result when it's convenient for them.

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,358 @@
...
+		err = proc.startTransport()
+	}
+	if err != nil {
+		proc.terminate(initialStatus, statusError, err)
+	}
+	return ctx, proc
+}
+
+func (self *authenticateeProcess) startTransport() error {
+	if err := self.transport.Start(); err != nil {
```
vladimirvivien: Won't **self.transport.Start()** bock here ? If so, the else{} block here will only be reached upon invoking **transport.Stop()** from another goroutine no? ... which may mean that the call to **self.transport.Stop()** in the else{} block would be unnecessary no?

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,358 @@
...
+	client    upid.UPID // pid of the client we're attempting to authenticate
+	handler   callback.Handler
+	transport messenger.Messenger // mesos communications transport
+}
+
+type transportFactory interface {
+	makeTransport() messenger.Messenger
+}
+
+type transportFactoryFunc func() messenger.Messenger
```
vladimirvivien: I struggle a bit to understand why use of the adapter function pattern to create a transport factory here.  Is this to allow a future (yet unknown) transport ? if so doesn't the *Messenger* interface decouples the implementation already.

auth/sasl/authenticatee.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+	if err := self.transport.Send(ctx, &pid, message); err != nil {
+		self.terminate(status, statusError, err)
+	} else {
+		(&self.status).swap(status, statusStarting)
+	}
+}
+
+func (self *authenticateeProcess) mechanisms(ctx context.Context, from *upid.UPID, pbMsg proto.Message) {
+	status := statusFrom(ctx)
```
vladimirvivien: **status** is pulled from context to validate current status.  However, I don't see status being put back in the *context* from previous step (in *authenticate()*, for instance).  Is status being saved in *authenticateProcess.status* or is it put into *context* in between steps ?

auth/sasl/authenticatee.go
jdef: Start() is non-blocking:
https://github.com/mesos/mesos-go/blob/master/messenger/messenger.go#L139

auth/sasl/authenticatee.go
jdef: How about for mocking in unit tests?
https://github.com/jdef/mesos-go/blob/crammd5/auth/sasl/authenticatee_test.go#L60

The factory is a func that produces a Messenger -- not that implements a Messenger. This pattern actually got rid of at least a dozen lines of duplicated code (between the test, and the actual implementation). I'm open to suggestions if you have a cleaner way to approach this.

Perhaps it's confusing because I named it "transport" even though it's an object of type Messenger?

auth/callback/interface.go
```diff
@@ -0,0 +1,28 @@
...
+
+type Unsupported struct {
+	Callback Interface
+}
+
+func (uc *Unsupported) Error() string {
+	return fmt.Sprintf("Unsupported callback <%T>: %v", uc.Callback, uc.Callback)
+}
+
+type Interface interface {
```
vladimirvivien: Curious, what's the advantage of using marker interface *Interface*  type vs *interface{}* type ? Still need type assertion i see.

auth/sasl/authenticatee.go
jdef: More function adapters involved here. Since I already wanted status near the `from` check I thought it would be cleaner to store it in context for the auth-process steps to use (there's only one `(&self.status).get()` per-step this way).

https://github.com/jdef/mesos-go/blob/crammd5/auth/sasl/authenticatee.go#L194

auth/callback/interface.go
jdef: I thought it would add clarity when looking at code outside the callback package:
```go
handler := callback.HandlerFunc(callbacks ...callback.Interface) {
    // ...
})
```
and then someone could review the docs for callback.Interface and understand how it's used in the scheme of things. Unfortunately I never wrote docs.

auth/sasl/authenticatee_test.go
```diff
@@ -0,0 +1,98 @@
...
+			case *callback.Interprocess:
+				c.Set(server, client)
+			default:
+				return &callback.Unsupported{c}
+			}
+		}
+		return nil
+	})
+	var transport *MockTransport
+	factory := transportFactoryFunc(func() messenger.Messenger {
```
vladimirvivien: Ok I see why transportFactory was made a function adapter type that returns *messenger.Messenger*.  Is this a matter of style since it would just as easily work using the *messenger.Messenger* interface.  Or is it indirection for yet-unknown impl ?

vladimirvivien:
I think there is a need for a README to provide a high-level overview of the auth plugin impl approach.  That will help down the road.

auth/sasl/authenticatee.go
vladimirvivien: My bad, I looked at http_transporter.go.  Never mind.

auth/sasl/authenticatee.go
vladimirvivien: Yes, I realized that it helped with testing.. see comment below.

auth/sasl/authenticatee.go
vladimirvivien: got it.

auth/sasl/authenticatee_test.go
jdef: I'm not sure I understand your suggestion. I used a factory function because I want a new instance of the messenger implementation for each new authentication attempt. If I instantiated the messenger once and just kept using that then, yes, I could get rid of the factory function. This is modeled on the design used in the Mesos native authenticatee, which spawns a new process (get's a new authentication PID) for each.

What's somewhat interesting to think about is to have the authentication messenger share the same HTTP transport as the driver messenger. However, I wouldn't want the authentication messenger to Start() or Stop() the HTTP transport since the driver expects to control the lifecycle of the transport that it created. But that way there'd be only a single HTTP listener handling traffic between the framework and the scheduler. It could probably be done fairly simply creating a SharedTransport implementation that wraps an existing transport, turns Start() into a noop, and re-implements Stop() to Uninstall the handlers registered via the wrapper (but that's it, it wouldn't actually stop the wrapped transport).

auth/sasl/authenticatee.go
jdef: This was implemented.

jdef:
@vladimirvivien thanks for the review!

PR #58, opened 2015-01-19
Updated mesos proto buffer to the latest version

silenteh:
This allows mesos-go to support the ContainerInfo property in both the
ExecutorInfo and TaskInfo and therefore allows to use the default mesos
executor to start Docker containers


jdef:
Looks like the changes from this PR have already merged in from other, parallel PRs. Thanks for submitting!

PR #60, opened 2015-01-20
use framework's hostname to initialize scheduler's UPID hostname

naelyn:
Frameworks do not quite expose themselves properly to the Mesos Master unless UPID.Host is set to match the listen address (conveniently FrameworkInfo.Hostname).


naelyn:
#69 fixes this, so closing

PR #66, opened 2015-01-30
Update README.md

dajac:
Correct typo.


vladimirvivien:
@dajac Thanks for the patch.

Issue #65, opened 2015-01-29
port 80


razic:
tried to run a http server in the background alongside the driver (separate go routine) but port 80 seems to be in use???

vladimirvivien:
@razic thanks for reporting. This may be related to #59.

PR #69, opened 2015-02-04
add ip/host scheduler flags

### LINKED ISSUE #59: Hostname/port Configuration for SchedulerDriver
Add the ability to specify the hostname and port the Scheduler driver should bind and advertise to the master.  See #52 for backstory.

razic:
creating this pull request for discussion.

you may want me to change a couple things.

fixes #59, #60, and #65

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -45,6 +45,10 @@ const (
 var (
 	authProvider = flag.String("mesos_authentication_provider", sasl.ProviderName,
 		fmt.Sprintf("Authentication provider to use, default is SASL that supports mechanisms: %+v", mech.ListSupported()))
+	schedulerHost = flag.String("mesos_scheduler_ip", "0.0.0.0",
+		fmt.Sprintf("TCP4 Address for the Scheduler to bind"))
+	schedulerPort = flag.String("mesos_scheduler_port", "0",
+		fmt.Sprintf("TCP Port for the Scheduler to bind"))
 )
 
 // Concrete implementation of a SchedulerDriver that connects a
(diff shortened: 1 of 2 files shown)


scheduler/scheduler.go
```diff
@@ -45,6 +45,10 @@ const (
 var (
 	authProvider = flag.String("mesos_authentication_provider", sasl.ProviderName,
 		fmt.Sprintf("Authentication provider to use, default is SASL that supports mechanisms: %+v", mech.ListSupported()))
+	schedulerHost = flag.String("mesos_scheduler_ip", "0.0.0.0",
+		fmt.Sprintf("TCP4 Address for the Scheduler to bind"))
```
fmpwizard: this gives a `go vet` error, same for two lines below from here

razic: @fmpwizard sorry, i'm not familiar with go vet. i usually just use https://github.com/fatih/vim-go which i think is gofmt. let me see if i can fix... can you paste the error?

razic: also i ran `go vet scheduler/scheduler.go` and it exited with 0 and no stdout.

fmpwizard: you can run go vet from the terminal, it checks for things that "could" be errors, but it is not correct 100% of the time. More info here https://golang.org/cmd/go/#hdr-Run_go_tool_vet_on_packages
I actually didn't run go vet on this file but the error you get is because `Sprintf` expects a format string and then the value(s) to use, in this case you are not passing any variables to use, so you can simply use `fmt.Println()`

fmpwizard: @razic ^ (forgot to include your nick)

fmpwizard: hm, I just run it and you are right, it doesn't bring the error, strange as I thought it should, ok, sorry for the noise

razic: actually `fmt.Println()` writes to STDOUT and does not return a string. i think you mean `fmt.Sprint`

fmpwizard: now that you bring that up, you can just put the string there, no need for any `Sprint*`

razic: haha its so funny how i didn't even notice that until you mentioned it.

vladimirvivien: @razic  @fmpwizard thanks for pointing out go vet.  Open a separate issue to track go vet issues.

vladimirvivien:
The code looks good and passing the tests.  The flag approach works with no disruption to the code base.  Merged.

razic:
:+1:

PR #70, opened 2015-02-06
Grab bag of reliability changes

jdef:
* https://github.com/mesosphere/kubernetes-mesos/issues/145
* https://github.com/mesosphere/kubernetes-mesos/issues/146
* https://github.com/mesosphere/kubernetes-mesos/issues/147
* .. and refactored unit tests to use a "suite" to isolate global test variables

PR #73, opened 2015-02-08
Implement framework reregistration

naelyn:
This backports C++ Mesos framework re-registration behavior into the Go native bindings.

Without this change frameworks are incapable of actually failing over.


jdef:
you may be interested in the "reliable registration" that I attempted to port from the native bindings as part of PR #70

vladimirvivien:
@naelyn #70 prepares the ground for adding master detection to align with cpp version.

jdef:
thanks for the PR. similar changes have already been merged to resolve this. if there's still a problem with re-registration please open a new issue.

PR #80, opened 2015-02-12
Integrating Scheduler with the Zookeeper Lead Detector

vladimirvivien:
To do 
- [x] Integration of scheduler.go with detector package
- [x] Fix detector.Standalone to return masterinfo instance.
- [x] Test NewSchedulerDriver with plain master value
- [x] Fix detector.factory to detect master value that starts with "master@" 
- [ ] Test NewScheduelerDriver with PID value.
- [ ] Test NewSchedulerDriver with master value as "zk://"
- [ ] Test scheduler master detection change behavior (ensure disconnection, re-connection)
- [ ] Test re-authentication when a new master is detected.


scheduler/scheduler.go
```diff
@@ -187,9 +198,59 @@ func (driver *MesosSchedulerDriver) init() error {
...
+
+	if master != nil {
+		log.Infof("New master %s detected\n", master.String())
+
+		pid, err := upid.Parse(master.GetPid())
+		if err != nil {
+			panic("Unable to parse Master's PID value.") // this should not happen.
+		}
+
+		driver.MasterPid = pid // save for downstream ops.
```
jdef: locking needed around this (and related reads) if other ops are accessing it concurrently?

scheduler/scheduler.go
```diff
@@ -187,9 +198,59 @@ func (driver *MesosSchedulerDriver) init() error {
...
+		log.Info("Ignoring master change because the driver is aborted.")
+		return
+	}
+
+	// Reconnect every time a master is dected.
+	if driver.Connected() {
+		log.V(3).Info("Disconnecting scheduler.")
+		driver.Scheduler.Disconnected(driver)
+	}
+
```
jdef: probably need to set authenticated=false

scheduler/scheduler.go
```diff
@@ -187,9 +198,59 @@ func (driver *MesosSchedulerDriver) init() error {
...
+				driver.lock.Lock()
+				defer driver.lock.Unlock()
+				driver.authenticated = true
+			}()
+
+		} else {
+			log.Infoln("No credentials were provided. " +
+				"Attempting to register scheduler without authentication.")
+			// register framework
+			go driver.doReliableRegistration(float64(registrationBackoffFactor))
```
jdef: when is registration triggered for authenticated frameworks?

scheduler/scheduler.go
```diff
@@ -97,6 +99,7 @@ type MesosSchedulerDriver struct {
 	updates         map[string]*mesos.StatusUpdate // Key is a UUID string.
 	tasks           map[string]*mesos.TaskInfo     // Key is a UUID string.
 	credential      *mesos.Credential
+	authenticated   bool
```
jdef: who reads this?

scheduler/scheduler.go
```diff
@@ -150,23 +153,31 @@ func NewMesosSchedulerDriver(
...
 
-	if m, err := upid.Parse("master@" + master); err != nil {
+	if md, err := detector.New(master); err != nil {
 		return nil, err
 	} else {
-		driver.MasterPid = m
+		driver.masterDetector = md
 	}
 
+	// if m, err := upid.Parse("master@" + master); err != nil {
```
jdef: remove dead code

scheduler/scheduler.go
```diff
@@ -187,9 +198,59 @@ func (driver *MesosSchedulerDriver) init() error {
...
+
+		pid, err := upid.Parse(master.GetPid())
+		if err != nil {
+			panic("Unable to parse Master's PID value.") // this should not happen.
+		}
+
+		driver.MasterPid = pid // save for downstream ops.
+
+		if driver.credential != nil {
+			if err := driver.authenticate(); err != nil {
```
jdef: authentication should probably not be done synchronously in the detector handler since it could potentially block and we really should try to avoid blocking for long in callbacks like this.

scheduler/scheduler.go
vladimirvivien: Not sure yet, if not needed, I will remove.

scheduler/scheduler.go
vladimirvivien: Agreed.

scheduler/scheduler.go
```diff
@@ -150,23 +153,31 @@ func NewMesosSchedulerDriver(
...
-	if m, err := upid.Parse("master@" + master); err != nil {
+	if md, err := detector.New(master); err != nil {
 		return nil, err
 	} else {
-		driver.MasterPid = m
+		driver.masterDetector = md
 	}
 
+	// if m, err := upid.Parse("master@" + master); err != nil {
+	// 	return nil, err
```
vladimirvivien: Ok.

jdef:
closing this out, mesos-go/zk2 branch now has these changes

PR #83, opened 2015-02-14
Allowing executor to listen on any port and bind to all network interfaces

### LINKED ISSUE #72: Executor startup broken
Hi,

I just pulled master, and saw that my executor doesn't start up anymore. This executor used to startup until a few weeks back. 

E0208 02:08:22.506744   15178 http_transporter.go:128] HTTPTransporter failed to listen: listen tcp4: unknown port tcp4/

E0208 02:08:22.507236   15178 messenger.go:150] Failed to start messenger: listen tcp4: unknown port tcp4/

E0208 02:08:22.507247   15178 executor.go:388] Failed to start executor: listen tcp4: unknown port tcp4/

@vladimirvivien Do you have any suggestions?

diptanu:
This fixes #72 

@vladimirvivien @nqn This will allow the executor to listen on all interfaces and on any ports. I am not aware of any way it's possible to make an executor listen on a particular port and ip and have the slave communicate with it. At least, for Java based executors setting ip and port for executor isn't required.

Please let me know if there is any work around or merge this as soon as possible since this is kind of a blocker.

### DIFF
--- a/messenger/http_transporter.go
+++ b/messenger/http_transporter.go
@@ -182,13 +182,6 @@ func (t *HTTPTransporter) Listen() error {
 	host := t.upid.Host
 	port := t.upid.Port
 
-	if host == "" {
-		host = "0.0.0.0"
-	}
-
-	if port == "" {
-		port = "0"
-	}
 	// NOTE: Explicitly specifies IPv4 because Libprocess
 	// only supports IPv4 for now.
 	ln, err := net.Listen("tcp4", net.JoinHostPort(host, port))
--- a/messenger/http_transporter.go
+++ b/messenger/http_transporter.go
@@ -181,6 +181,14 @@ func (t *HTTPTransporter) Install(msgName string) {
 func (t *HTTPTransporter) Listen() error {
 	host := t.upid.Host
 	port := t.upid.Port
+
+	if host == "" {
+		host = "0.0.0.0"
+	}
+
+	if port == "" {
+		port = "0"
+	}
 	// NOTE: Explicitly specifies IPv4 because Libprocess
 	// only supports IPv4 for now.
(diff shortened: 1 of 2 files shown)


messenger/http_transporter.go
```diff
@@ -181,6 +181,14 @@ func (t *HTTPTransporter) Install(msgName string) {
 func (t *HTTPTransporter) Listen() error {
 	host := t.upid.Host
 	port := t.upid.Port
+
+	if host == "" {
+		host = "0.0.0.0"
+	}
+
+	if port == "" {
+		port = "0"
```
fmpwizard: is port 0 really valid? shouldn't the default be 5051 for slaves and 5050 for master nodes?

vladimirvivien:
@diptanu thanks for the pr.  However, I would suggest to handle this the same way PR #69.  Simply put, instead of changing the http-transporter, provide sensible defaults via flags, like mesos_executor_ip and mesos_executor_port.  That way,  you can override those value easily.

messenger/http_transporter.go
diptanu: @fmpwizard So this is not for the slave, but for custom executors. Custom executors need to listen on a port for connecting to the slaves. And port 0 allows the custom executor to listen on any ephemeral port which is free.

diptanu:
@vladimirvivien Thanks for the review, updated the PR with flags.

messenger/http_transporter.go
fmpwizard: ah ok, thanks for the clarification

diptanu:
@vladimirvivien Any update on this?

jdef:
We're going to strip flags like this out from the library, they are really the responsibility of the scheduler and executor not the bindings library.

Please see commit 526c340

Ongoing work in the zk2 branch -- feedback welcome.

PR #97, opened 2015-02-24
Use examples build target when invoking the all target as it calls both test-framework and test-executor already

chrisjs:
trivial; should have been done in prior PR but just noticed it..


vladimirvivien:
Tested and looks good.  Thanks @chrisjs

PR #101, opened 2015-02-27
Rename test-executor and test-framework references to example-executor and example-scheduler to be more consistent with example naming

chrisjs:
nit-picky but having the "test" prefix for examples always seemed a little inconsistent to me

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,8 +91,8 @@ _testmain.go
 junit_*
 
 # mesos-go specific
-examples/test-framework
-examples/test-executor
+examples/example-scheduler
+examples/example-executor
 
 #webide
 .c9
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,2 +1,2 @@
 From busybox:ubuntu-14.04 
-ADD ./example_executor/example_executor /bin/test-executor
+ADD ./example_executor/example_executor /bin/example-executor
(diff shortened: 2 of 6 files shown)


fmpwizard:
now that you are renaming this, what do you think of also renaming `example-framework` to `example-scheduler` ?

jdef:
In the mesos/src/examples tree, the referenced java and python example schedulers and executors are actually called TestFramework/TestExecutor, and test_framework/test_executor. I imagine that this project does the same for the sake of consistency with the mesos project. But I agree that it's not very go-ish.

cc @vladimirvivien

chrisjs:
@fmpwizard - seems reasonable, will update PR providing its agreed the name changes as a whole are acceptable

@jdef - yes I've noticed the same - yet while being consistent with the mesos project, possibly those names (test-*) are historical artifacts and the java/python ones have inherited the same naming conventions. in the end, IMO example-* seems to better describe the files then test-* and given they are examples i don't think it would be that bad to deviate in the naming for consistency. Also structs, etc inside these files are named using "example" as well

chrisjs:
updated PR to include the suggestion from @fmpwizard to replace framework with scheduler

fmpwizard:
Thanks for the updates @chrisjs , hopefully one of the maintainers will agree to merge this in

chrisjs:
@vladimirvivien :+1:

jdef:
LGTM - thanks for the PR

PR #111, opened 2015-04-10
zk api extension

### LINKED ISSUE #110: expose API for listing all currently detected mesos masters, not just the leader
/cc @kozyraki

jdef:
/cc @kozyraki

resolves #110 

TODO:
- [x] unit tests

### DIFF
--- a/detector/interface.go
+++ b/detector/interface.go
@@ -27,6 +27,18 @@ type MasterChanged interface {
 	OnMasterChanged(*mesos.MasterInfo)
 }
 
+// optional interface that, if implemented by the same object as implements
+// MasterChanged, will receive an additional callbacks independently of
+// leadership changes. it's possible that, as a result of a leadership change,
+// both the OnMasterChanged and UpdatedMasters callbacks would be invoked.
+// the additional callback, UpdatedMasters, is useful to clients that wish to know
+// the entire set of Mesos masters currently running.
+// **NOTE:** Detector implementations are not required to support this optional
+// interface. Please RTFM of the detector implementation that you want to use.
+type AllMasters interface {
+	UpdatedMasters([]*mesos.MasterInfo)
+}
+
 // func/interface adapter
 type OnMasterChanged func(*mesos.MasterInfo)
 
--- a/detector/interface.go
+++ b/detector/interface.go
@@ -27,15 +27,18 @@ type MasterChanged interface {
 	OnMasterChanged(*mesos.MasterInfo)
 }
 
-// optional interface that, if implemented by the same object as implements
-// MasterChanged, will receive an additional callbacks independently of
-// leadership changes. it's possible that, as a result of a leadership change,
-// both the OnMasterChanged and UpdatedMasters callbacks would be invoked.
-// the additional callback, UpdatedMasters, is useful to clients that wish to know
-// the entire set of Mesos masters currently running.
+// AllMasters defines an optional interface that, if implemented by the same
+// struct as implements MasterChanged, will receive an additional callbacks
+// independently of leadership changes. it's possible that, as a result of a
+// leadership change, both the OnMasterChanged and UpdatedMasters callbacks
+// would be invoked.
+//
 // **NOTE:** Detector implementations are not required to support this optional
 // interface. Please RTFM of the detector implementation that you want to use.
 type AllMasters interface {
+	// UpdatedMasters is invoked upon a change in the membership of mesos
+	// masters, and is useful to clients that wish to know the entire set
+	// of Mesos masters currently running.
 	UpdatedMasters([]*mesos.MasterInfo)
 }
 
--- a/detector/zoo/detect.go
+++ b/detector/zoo/detect.go
@@ -112,8 +112,12 @@ func (md *MasterDetector) childrenChanged(zkc *Client, path string, obs detector
 		return
 	}
 
-	topNode := selectTopNode(list)
+	md.notifyMasterChanged(path, list, obs)
+	md.notifyAllMasters(path, list, obs)
+}
 
+func (md *MasterDetector) notifyMasterChanged(path string, list []string, obs detector.MasterChanged) {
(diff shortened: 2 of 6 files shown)


detector/interface.go
```diff
@@ -27,6 +27,18 @@ type MasterChanged interface {
 	OnMasterChanged(*mesos.MasterInfo)
 }
 
+// optional interface that, if implemented by the same object as implements
```
tsenart: If we follow Godoc conventions, documentation is gonna be nice. Let's start this comment with the name of the entity it describes. `AllMasters defines ...`

detector/interface.go
```diff
@@ -27,6 +27,18 @@ type MasterChanged interface {
...
+// MasterChanged, will receive an additional callbacks independently of
+// leadership changes. it's possible that, as a result of a leadership change,
+// both the OnMasterChanged and UpdatedMasters callbacks would be invoked.
+// the additional callback, UpdatedMasters, is useful to clients that wish to know
+// the entire set of Mesos masters currently running.
+// **NOTE:** Detector implementations are not required to support this optional
+// interface. Please RTFM of the detector implementation that you want to use.
+type AllMasters interface {
+	UpdatedMasters([]*mesos.MasterInfo)
```
tsenart: In a parallel universe, letting the user subscribe to framework events through a channel would be ace.

detector/zoo/detect.go
```diff
@@ -124,21 +128,51 @@ func (md *MasterDetector) childrenChanged(zkc *Client, path string, obs detector
...
-			log.Errorf("unable to retrieve leader data: %v", err.Error())
-			return
+		var err error
+		if masterInfo, err = md.pullMasterInfo(path, topNode); err != nil {
+			log.Errorln(err.Error())
 		}
+	}
+	log.V(2).Infof("detected master info: %+v", masterInfo)
+
+	//TODO(jdef) recover from possible client callback panic here, so that it doesn't
```
tsenart:
```go
defer func() {
  if err := recover(); err != nil {
    log.Errorln(err)
  }
}()
obs.OnMasterChanged(masterInfo)

Comment (vladimirvivien) [inline detector/zoo/detect.go]: @tsenart Why the explicit panic recovery ?

File: detector/interface.go (hunk 2 above)
Comment (jdef) [inline detector/interface.go]: I thought about implementing that. I'd like to avoid the case where the detector tries to publish an event to a chan but the client has either disappeared or is otherwise not reading from the chan fast enough. So this seemed like a pretty low overhead approach that allows us to avoid that.

File: detector/zoo/detect.go (hunk 3 above)
Comment (jdef) [inline detector/zoo/detect.go]: don't trust clients

File: detector/interface.go (hunk 2 above)
Comment (tsenart) [inline detector/interface.go]: Wouldn't a non-blocking send solve the problem of readers disappearing or being slow?
```go
select {
case eventch <- ev:
default:
    // reader is too slow, drop event
}
```
If we want to support some lag from readers, we can tune the channel buffer to some value bigger than zero.

jdef: It would, except they would miss events if they were too slow. Yes we could
use a buffered chan .. how big is big enough if a client has disappeared?

tsenart: If a client disappears, I would assume it would receive the whole state
when coming back.

On Saturday, May 30, 2015, James DeFelice <notifications@github.com> wrote:

> In detector/interface.go
> <https://github.com/mesos/mesos-go/pull/111#discussion_r31379528>:
>
> > @@ -27,6 +27,18 @@ type MasterChanged interface {
> >  	OnMasterChanged(*mesos.MasterInfo)
> >  }
> >
> > +// optional interface that, if implemented by the same object as implements
> > +// MasterChanged, will receive an additional callbacks independently of
> > +// leadership changes. it's possible that, as a result of a leadership change,
> > +// both the OnMasterChanged and UpdatedMasters callbacks would be invoked.
> > +// the additional callback, UpdatedMasters, is useful to clients that wish to know
> > +// the entire set of Mesos masters currently running.
> > +// **NOTE:** Detector implementations are not required to support this optional
> > +// interface. Please RTFM of the detector implementation that you want to use.
> > +type AllMasters interface {
> > +	UpdatedMasters([]*mesos.MasterInfo)
>
> It would, except they would miss events if they were too slow. Yes we
> could use a buffered chan .. how big is big enough if a client has
> disappeared?
> … <#14da45e9e562b843_>
> On Sat, May 30, 2015 at 4:09 AM, Tomás Senart <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote: In
> detector/interface.go <
> https://github.com/mesos/mesos-go/pull/111#discussion_r31378872>: > @@
> -27,6 +27,18 @@ type MasterChanged interface { >
> OnMasterChanged(*mesos.MasterInfo) > } > > +// optional interface that, if
> implemented by the same object as implements > +// MasterChanged, will
> receive an additional callbacks independently of > +// leadership changes.
> it's possible that, as a result of a leadership change, > +// both the
> OnMasterChanged and UpdatedMasters callbacks would be invoked. > +// the
> additional callback, UpdatedMasters, is useful to clients that wish to know
> > +// the entire set of Mesos masters currently running. > +// **NOTE:**
> Detector implementations are not required to support this optional > +//
> interface. Please RTFM of the detector implementation that you want to use.
> > +type AllMasters interface { > + UpdatedMasters([]*mesos.MasterInfo)
> Wouldn't a non-blocking send solve the problem of readers disappearing or
> being slow? select {case eventch <- ev:default: // reader is too slow, drop
> event } If we want to support some lag from readers, we can tune the
> channel buffer to some value bigger than zero. — Reply to this email
> directly or view it on GitHub <
> https://github.com/mesos/mesos-go/pull/111/files#r31378872>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mesos/mesos-go/pull/111/files#r31379528>.
>


-- 
Sent from Gmail Mobile

PR #113, opened 2015-04-17
tests: Fail fast on drivers' construction errors

tsenart:
Running tests locally with go test ./... was causing several panics due
to failing DNS queries for the locally set hostname ("lookup senart: no such host")

The assertion library being used wasn't failing fast in presence of such errors
being returned by driver constructors. This resulted in nil drivers
being accessed and hence causing panics.

This change set makes these constructors abort the tests immediately instead of
delaying the error reporting.

### DIFF
--- a/executor/executor_intgr_test.go
+++ b/executor/executor_intgr_test.go
@@ -19,13 +19,6 @@
 package executor
 
 import (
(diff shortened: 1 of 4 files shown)


jdef:
LGTM - thanks for the PR

PR #116, opened 2015-04-26
Log warning if no username can be obtained (and none is given).

ZeissS:
I crosscompiled the example_scheduler on my macbook and copied it over to the playa-vm. Here the `user.Current()` call failed (because its not support in cross-compiled binaries in go) __silently__.

This PR logs a warning if no username is given in the `DriverConfig` and is failed to be obtained by `user.Current()`.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -161,7 +161,7 @@ func NewMesosSchedulerDriver(config DriverConfig) (initializedDriver *MesosSched
 			if err != nil {
 				log.Warningln("Failed to obtain username: %v", err)
 			} else {
-				log.Warningln("Failed to obtain username.")
+				log.Warningf("Failed to obtain username.")
 			}
 			framework.User = proto.String("")
 		} else {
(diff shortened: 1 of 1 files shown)


ZeissS:
NOTE: mesos itself (=framework registration) seem to have worked fine w/o a username, but executing tasks in the example_executer failed.

scheduler/scheduler.go
```diff
@@ -158,6 +158,11 @@ func NewMesosSchedulerDriver(config DriverConfig) (initializedDriver *MesosSched
 	if framework.GetUser() == "" {
 		user, err := user.Current()
 		if err != nil || user == nil {
+			if err != nil {
+				log.Warningln("Failed to obtain username: %v", err)
```
jdef: This should probably be Warningf

ZeissS: Right, sorry :) last minute changes to make the output better.

jdef:
Thanks for this

PR #117, opened 2015-04-26
Support PID hosts with IPs

ZeissS:
While playing around with https://github.com/mesosphere/playa-mesos and mesos-go I failed to connect the example framework to the master. The master identified itself as `master@127.0.1.1:5050` which failed the parsing logic in `detector/factory.go`.

This PR fixes `CreateMasterInfo()` to support PIDs with IPs as the hostname.

### DIFF
--- a/detector/factory.go
+++ b/detector/factory.go
@@ -126,7 +126,9 @@ func CreateMasterInfo(pid *upid.UPID) *mesos.MasterInfo {
 	}
 	//TODO(jdef) what about (future) ipv6 support?
 	var ipv4 net.IP
-	if addrs, err := net.LookupIP(pid.Host); err == nil {
+	if ipv4 = net.ParseIP(pid.Host); ipv4 != nil {
+		// Intentionally left empty. ipv4 is successfully parsed, so we can skip the other versions
+        } else if addrs, err := net.LookupIP(pid.Host); err == nil {
 		for _, ip := range addrs {
 			if ip = ip.To4(); ip != nil {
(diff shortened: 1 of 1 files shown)


jdef:
Thanks for the PR! Can you please provide a unit test that covers this case?

ZeissS:
Dang. I started writing a test but then noticed that it works fine on my mac w/o the change. Digged a bit further and noticed that this is another cross-compile problem (see [0]).

I am not sure it makes sense to merge this then. Feel free to close. Otherwise I am not sure how to provide a test-case that actually fails if my change is not applied.

[0] https://groups.google.com/forum/#!topic/golang-nuts/nVqoIKHdZ6s

jdef:
Thanks for digging into this. Will merge as-is.

drewkerrigan:
Commenting here in preparation for an upcoming PR:

```
if ipv4 = net.ParseIP(pid.Host); ipv4 != nil {
...
packedip := binary.BigEndian.Uint32(ipv4) // network byte order is big-endian
```

always results in `packedip` getting set to `0` because `ipv4` in this case isn't actually converted to the 4 byte representation [https://golang.org/pkg/net/#IP.To4](https://golang.org/pkg/net/#IP.To4)

Changing the line to:

```
if ipv4 = net.ParseIP(pid.Host).To4(); ipv4 != nil {
```

fixes the problem, but I am revamping `CreateMasterInfo` currently to handle ipv6 as well.

PR #118, opened 2015-04-28
only set upid.Host and upid.Port if Host is unset

antonlindstrom:
This commit checks if `upid.Host` is set in the http_transporter and sets it
to the value that it determines if unset.

The original problem was that I was trying to run a scheduler on 0.0.0.0 as
listening host and have OverrideHostname set, this should have led to the
master thinking that I was sending messages from HostnameOverride, instead the
http_transporter function listen(), mutated the upid.Host and the master would
always look up 0.0.0.0.

The patch has been tested by running a scheduler against a Mesos cluster with
the scheduler bound to 0.0.0.0 and HostnameOverride in DriverConfig to an
external IP.

Thanks!

### DIFF
--- a/messenger/http_transporter_test.go
+++ b/messenger/http_transporter_test.go
@@ -2,6 +2,7 @@ package messenger
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"regexp"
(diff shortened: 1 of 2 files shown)


jdef:
Thanks for the PR! Can you also write up a quick unit test?

antonlindstrom:
@jdef Thanks! I added a simple test that covered my case, it's fairly simple though.

messenger/http_transporter_test.go
```diff
@@ -266,6 +267,34 @@ func TestTransporterStartAndStop(t *testing.T) {
...
+
+	// override the upid.Host with this listener IP
+	addr := net.ParseIP("127.0.1.1")
+
+	// setup receiver (server) process
+	uPid, err := upid.Parse(fmt.Sprintf("%s@%s", serverId, serverAddr))
+	assert.NoError(t, err)
+	receiver := NewHTTPTransporter(uPid, addr)
+
+	errch := receiver.Start()
```
jdef: - why call Start/Stop instead of listen(), which is the func that you really changed?
- this is great for testing the code as it was *before* your changes. how about adding a test case that stresses the code you added to HTTPTransporter?

messenger/http_transporter_test.go
```diff
@@ -266,6 +267,34 @@ func TestTransporterStartAndStop(t *testing.T) {
 	}
 }
 
+func TestMutadedHostUPid(t *testing.T) {
```
jdef: s/Mutaded/Mutated

antonlindstrom:
@jdef Thanks for the feedback! I've updated the tests to include a test where the upid host and port are empty. I think that should cover what I actually changed.

Thanks!

messenger/http_transporter_test.go
```diff
@@ -266,6 +267,66 @@ func TestTransporterStartAndStop(t *testing.T) {
...
+
+	// override the upid.Host with this listener IP
+	addr := net.ParseIP("127.0.1.1")
+
+	receiver := NewHTTPTransporter(uPid, addr)
+
+	err = receiver.listen()
+	assert.NoError(t, err)
+
+	// This should be the port that overrides as uPid.Host is empty
```
jdef: s/port/host

jdef:
Minor doc error. Otherwise LGTM

antonlindstrom:
Updated, thanks!

PR #125, opened 2015-05-15
Fix errata

jdef:


betawaffle:
What is preventing this from being merged?

jdef:
I think it just needs review

/cc @tsenart

jdef:
... and possibly a rebase

jdef:
just rebased and pushed

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
 
 	suite.False(driver.Stopped())
 	suite.Equal(mesos.Status_DRIVER_RUNNING, driver.Status())
+	driver.connected = true // pretend that we're already registered
```
tsenart: Is it possible to use a public method to get to this state instead?

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
...
 
 	driver.Stop(false)
+
+	msg := messenger.lastMessage
+	suite.NotNil(msg)
+	_, isUnregMsg := msg.(proto.Message)
+	suite.True(isUnregMsg, "expected UnregisterFrameworkMessage instead of %+v", msg)
+
+	time.Sleep(time.Millisecond * 1)
```
tsenart: I wouldn't be able to tell why we're sleeping here.

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
...
+}
+
+func (suite *SchedulerTestSuite) TestSchdulerDriverStop_WithFailover() {
+	// Set expections and return values.
+	messenger := &msgTracker{MockedMessenger: messenger.NewMockedMessenger()}
+	messenger.On("Start").Return(nil)
+	messenger.On("UPID").Return(&upid.UPID{})
+	messenger.On("Send").Return(nil)
+	messenger.On("Stop").Return(nil)
+	messenger.On("Route").Return(nil)
```
tsenart: When working on the next mesos-go bindings, I really want to avoid this sort of temporal coupling.

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
...
+	driver.messenger = messenger
+	suite.True(driver.Stopped())
+
+	go func() {
+		stat, err := driver.Run()
+		suite.NoError(err)
+		suite.Equal(mesos.Status_DRIVER_STOPPED, stat)
+	}()
+
+	time.Sleep(time.Millisecond * 1)
```
tsenart: Same comment as above.

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
...
+	// Set expections and return values.
+	messenger := &msgTracker{MockedMessenger: messenger.NewMockedMessenger()}
+	messenger.On("Start").Return(nil)
+	messenger.On("UPID").Return(&upid.UPID{})
+	messenger.On("Send").Return(nil)
+	messenger.On("Stop").Return(nil)
+	messenger.On("Route").Return(nil)
+
+	driver := newTestSchedulerDriver(suite.T(), NewMockScheduler(), suite.framework, suite.master, nil)
+	driver.messenger = messenger
```
tsenart: Same comment as above.

scheduler/scheduler_unit_test.go
```diff
@@ -356,8 +367,51 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStopOK() {
...
+	time.Sleep(time.Millisecond * 1)
+
+	suite.False(driver.Stopped())
+	suite.Equal(mesos.Status_DRIVER_RUNNING, driver.Status())
+	driver.connected = true // pretend that we're already registered
+
+	driver.Stop(true)
+
+	msg := messenger.lastMessage
+	suite.Nil(msg)
```
tsenart: It's not entirely clear to me why this msg should be nil.

scheduler/scheduler_unit_test.go
jdef: we don't really want a public method to set the driver state. it's definitely something that should be package-private.

jdef:
address review comments, @tsenart PTAL

scheduler/scheduler_unit_test.go
tsenart: What I really wanted to ask was if there is a way to reach the same state using only public methods instead of short-cutting and mutating internal state.

jdef: Probably, by simulating the registration stuff. Doing things this way keep
the overall scope of the test a bit more focused. I don't think mutating
internal state like this is such a huge deal for a narrowly focused unit
test. If this were a larger component or integration test then I'd go with
the public method approach in order to transition the project properly.

tsenart:
LGTM

PR #130, opened 2015-06-05
Fix races in executor, zoo and healthchecker packages

s-urbaniak:


tsenart:
While killing all these races down, @s-urbaniak also fixed some in our test library: https://github.com/stretchr/testify/pull/177

tsenart:
PTAL

tsenart:
LGTM

executor/executor.go
```diff
@@ -345,6 +350,7 @@ func (driver *MesosExecutorDriver) statusUpdateAcknowledgement(from *upid.UPID,
 
 	// Remove the corresponding update.
 	delete(driver.updates, uuid.String())
+	driver.lock.Unlock()
```
jdef: should we unlock, perhaps, after the next line that deletes from driver.tasks?

tsenart: SGTM

jdef:
minor nit, otherwise LGTM

PR #131, opened 2015-06-06
Some more CI improvements

tsenart:
This change set changes enables go fmt checking, go vet, race detector and building in Go 1.3, 1.4.2 and tip in Travis CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 EXAMPLES = examples
 
-.PHONY: format all go-clean pkg-build-install example-scheduler example-executor test
+.PHONY: format formatted all go-clean pkg-build-install test vet check \
+        example-scheduler example-executor
 
 all: go-clean pkg-build-install examples
 
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,6 @@
 EXAMPLES = examples
 
-PKG_PREFIX := github.com/mesos/mesos-go
-LIBS :=	\
-	auth \
-	auth/callback \
-	auth/sasl \
-	auth/sasl/mech \
-	auth/sasl/mech/crammd5 \
-	detector \
-	detector/zoo \
-	executor \
-	healthchecker \
-	mesosproto \
-	mesosutil \
-	mesosutil/process \
-	messenger \
-	scheduler \
-	upid
-
-.PHONY: format all go-clean pkg-build-install example-scheduler example-executor test test.v
+.PHONY: format all go-clean pkg-build-install example-scheduler example-executor test
 
 all: go-clean pkg-build-install examples
 
@@ -26,7 +8,7 @@ go-clean:
 	go clean
 
 pkg-build-install:
-	go install -v ${LIBS:%=./%}
+	go install -v ./...
 
 examples: example-scheduler example-executor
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -39,10 +21,13 @@ example-executor:
...
+formatted:
+	! gofmt -s -d . 2>&1 | read
 
 vet:
-	go vet ${LIBS:%=$(PKG_PREFIX)/%}
+	go vet ./...
 
-test test.v:
-	flags=""; test "$@" != "test.v" || flags="-test.v"; pkg="${TEST}"; test -n "$$pkg" || pkg="${LIBS:%=$(PKG_PREFIX)/%}"; go test $$pkg $$flags
+test: formatted vet
```
jdef: please don't remove the ability to specify a single package for testing via `make test TEST=...`. it's super-convenient. as is allowing a developer to choose between `test` and `test -v` output.

jdef: also, why should `test` depend upon `vet`? you can add an extra call to `vet` in the Travis YML but I wouldn't force it to run every time that `test` does. same with `formatted`.

tsenart: Why would we provide such functionality through the Makefile when it's implemented by the Go tool? If a Go developer wants to run any of these things independently he'll intuitively pick the Go tool.

tsenart: However, before committing any code, I'd like to have a single command to verify that I didn't break anything. That includes tests, vet, formatting and race detection.

jdef: > However, before committing any code, I'd like to have a single command...

Could we call it something like `make check`? That seems suitably generic and doesn't confuse the historical use of `make test`. As a dev I expect that `make test` **only** runs tests.

jdef: > Why would we provide such functionality through the Makefile

Looking again I see that this `make test TEST=...` impl doesn't really do what I'd want it to anyway. I really want to avoid having to type `go test -v github.com/mesos/mesos-go/detector/zoo`; I'd much rather type `make test TESTS=detector/zoo`. I don't think the current impl supports that notation though so we can scrap it.

I do like `test.v` because then I can see a breakdown of how long each test case takes. Given that it's pretty easy to run `go test -v ./...` we can get rid of `test.v` (though it does keep my fingers happier).

tsenart: Fair point regarding the expectation for `make test`. Even though `make check` has also historical uses, I agree that it's more appropriate in this case.

There's no need to type `go test -v github.com/mesos/mesos-go/detector/zoo` since relative paths are accepted by the go tool, this would be enough: `go test -v ./detector/zoo`.

tsenart:
PTAL

Makefile
```diff
@@ -1,24 +1,6 @@
...
-	healthchecker \
-	mesosproto \
-	mesosutil \
-	mesosutil/process \
-	messenger \
-	scheduler \
-	upid
-
-.PHONY: format all go-clean pkg-build-install example-scheduler example-executor test test.v
+.PHONY: format all go-clean pkg-build-install example-scheduler example-executor test
```
jdef: can we add `formatted` and `check` here?

jdef:
minor nit, otherwise LGTM

Makefile
tsenart: Oversaw that. Thanks.

tsenart:
Merged.

PR #132, opened 2015-06-07
messenger: Make tests pass on Mac OSX

tsenart:
This commit makes tests pass by using more commonly available loopback
addresses and explicitly requesting random port assigment with port 0.


jdef:
lgtm :ship:

PR #133, opened 2015-06-07
Bound tests run time

tsenart:
This change set makes sets the upper bound on test run time to 60s and reduces run time of the upid package tests.


upid/upid_test.go
```diff
@@ -29,17 +20,10 @@ func TestUPIDParse(t *testing.T) {
...
-		ra := generateRandomString()
-		u, err = Parse(ra)
-		if u != nil {
-			println(ra)
-		}
-		assert.Nil(t, u)
-		assert.Error(t, err)
-	}
-
+	assert.Nil(t, quick.Check(func(s string) bool {
```
jdef: :+1:

jdef:
lgtm :ship:

PR #134, opened 2015-06-07
Read gofmt diff to dummy variable

tsenart:
This commit fixes a Travis CI issue where their sandbox shell complains
if `read` doesn't receive any arguments.


jdef:
lgtm :ship:

PR #136, opened 2015-06-12
fix: messenger.ForHostname fails if the hostname is an IP

### LINKED ISSUE #135: messenger.ForHostname fails if the hostname is an IP
It fails with:

```
service.go:403] failed to create executor driver: lookup 192.168.33.67: invalid domain name
```

Compare mesosphere/kubernetes-mesos#343

jdef:
- fixes #135

### DIFF
--- a/messenger/messenger.go
+++ b/messenger/messenger.go
@@ -76,41 +76,62 @@ type MesosMessenger struct {
 	tr                Transporter
 }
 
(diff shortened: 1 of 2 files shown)


sttts:
Fixed in my setup with this patch.

jdef:
:ship:

PR #137, opened 2015-06-12
fix race in auth/ssl test

jdef:


tsenart:
Tests are failing :(

jdef:
Indeed, though not the auth/sasl tests that this PR fixes.

tsenart:
LGTM :shipit:

PR #147, opened 2015-07-22
Build instructions now use GOPATH and build godep

ivucica:
Previously godep was fetched, but not built. Path to it was also not specified.


ivucica:
Please don't merge this. I did not manage to build using updated instructions after all.

PR #148, opened 2015-07-22
Build instructions now use GOPATH and build godep

ivucica:
Previously godep was fetched, but not built. Path to it was also not specified.

This also includes an added godep dependency on golang.org/x/text which was missing and breaking the build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,13 +30,15 @@ This is a very early version of the project.  Howerver, here is a list of things
 ## Build Instructions
 The following instructions is to build the code from `github`.The project uses the `GoDep` for dependency management.
 ```
-$ cd <go-workspace>/src/
-$ mkdir -p github.com/mesos
-$ cd github.com/mesos
+$ export GOPATH=<go-workspace>
+$ mkdir -p ${GOPATH}/src/github.com/mesos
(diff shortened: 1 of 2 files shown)


jaytaylor:
@ivucica Is `./godep` really a valid command? See: https://github.com/mesos/mesos-go/pull/148/files#diff-04c6e90faac2675aa89e2176d2eec7d8R40

ivucica:
@jaytaylor IIRC, godep binary got deployed into 'current directory', but I'll have to try this again (and I will probably rebase on current head).

README.md
```diff
@@ -30,13 +30,15 @@ This is a very early version of the project.  Howerver, here is a list of things
 ## Build Instructions
 The following instructions is to build the code from `github`.The project uses the `GoDep` for dependency management.
 ```
-$ cd <go-workspace>/src/
-$ mkdir -p github.com/mesos
-$ cd github.com/mesos
+$ export GOPATH=<go-workspace>
```
fmpwizard: I think this will end up confusing users, on the README, under the requirements, we have `A standard and working Go workspace setup`

This would mean the user is expected to already have a working `GOPATH`, if the user is advance enough to have multiple `GOPATH`, they won't even need to read build instructions.

PR #149, opened 2015-07-28
WIP: initial revision of streaming API and rough impl

jdef:
Prototype of an idea for a streaming interface. This proposal only defines streaming for the scheduler. It's easy to see how a similar approach could be taken for the executor.

Interested in feedback.

### DIFF
--- a/scheduler/streaming/adapter.go
+++ b/scheduler/streaming/adapter.go
@@ -1,0 +1,97 @@
+package streaming
+
+import (
+	mesos "github.com/mesos/mesos-go/mesosproto"
+	"github.com/mesos/mesos-go/scheduler"
+)
+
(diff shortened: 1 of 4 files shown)


tsenart:
I'll take a look at this tomorrow. Exciting times!

jdef:
Well, don't get too excited :p  It's basically a thin streaming API that wraps the existing protobuf implementation (and all the baggage that entails). What I was aiming for here:

* a simple chan-based streaming interface -- does this feel like idiomatic Go?
* simpler bootstrapping for clients -- one func gets you an object that provides `Events() <-chan`
* an interface that we could start playing with *today* before the underlying HTTP-streaming stuff is ready

jdef:
the core team is working on a prototype here: https://github.com/jimenez/mesoscon-demo

PR #150, opened 2015-07-30
Port libprocess PID parser

vladimirvivien:
More robust validation of uid string using Regexp

EDIT: Let's port libprocess parser instead: https://github.com/apache/mesos/blob/0.24.0/3rdparty/libprocess/src/pid.cpp#L84

### DIFF
--- a/upid/upid.go
+++ b/upid/upid.go
@@ -21,7 +21,12 @@ package upid
 import (
 	"fmt"
 	"net"
+	"regexp"
 	"strings"
+)
+
+var (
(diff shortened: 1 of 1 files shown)


tsenart:
How does libmesos do this?

jdef:
very carefully: https://github.com/apache/mesos/blob/0.24.0/3rdparty/libprocess/src/pid.cpp#L84

tsenart:
Let's port that method to Go instead of using a regex.

jdef:
:+1:

jdef:
closing this out due to old age. please re-open if/when interested.

PR #151, opened 2015-08-03
Switch from to code.google.com/p/go-uuid/uuid to github.com/pborman/uuid

eparis:
As code.google.com is going away someday


jdef:
lgtm

PR #154, opened 2015-08-07
Allow a different address published than binded.

tnachen:
Currently we only allow setting the binding address, which is what ip/port the messenger is going to bind to and be sending to the master as the UPID host.
However in some hosts the public ip differs than the private local ip that it can bind to, so allow publishing  a differnet ip can allow the master to connect back from a different public ip.

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -41,6 +41,7 @@ type DriverConfig struct {
 	HostnameOverride string                              // optional
 	BindingAddress   net.IP                              // optional
 	BindingPort      uint16                              // optional
+	PublishedAddress net.IP                              // optional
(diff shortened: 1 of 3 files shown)


jdef:
build fails:
```
executor/executor.go:85: not enough arguments in call to messenger.ForHostname
```

tnachen:
Sorry about that, pushed a fix now.

tnachen:
I think make check is failing on master already, with the same errors.

tnachen:
@jdef is tests failing for you too?

jdef:
yes, the tests have a slew of known issues. i think your changes look fine.

jdef:
:ship:

PR #164, opened 2015-09-09
added role as a parameter to the Go example framework

clintonskitson:
In order to take advantage of slaves offering specific resources for specific roles, the framework must be able to register as a certain role and submit jobs for resources with this role.  This PR adds the `--role` flag to the CLI.  All resources offered should match the `role` specified, and the tasks should be submitted with resources matching this `role`.

### DIFF
--- a/examples/example_scheduler.go
+++ b/examples/example_scheduler.go
@@ -56,6 +56,7 @@ var (
 	taskCount           = flag.String("task-count", "5", "Total task count to run.")
 	mesosAuthPrincipal  = flag.String("mesos_authentication_principal", "", "Mesos authentication principal.")
 	mesosAuthSecretFile = flag.String("mesos_authentication_secret_file", "", "Mesos authentication secret file.")
+	role                = flag.String("role", "*", "Role to login framework with")
 )
 
 type ExampleScheduler struct {
@@ -129,8 +130,8 @@ func (sched *ExampleScheduler) ResourceOffers(driver sched.SchedulerDriver, offe
 				SlaveId:  offer.SlaveId,
 				Executor: sched.executor,
 				Resources: []*mesos.Resource{
-					util.NewScalarResource("cpus", CPUS_PER_TASK),
-					util.NewScalarResource("mem", MEM_PER_TASK),
+					util.NewScalarResource("cpus", CPUS_PER_TASK, util.Role(role)),
+					util.NewScalarResource("mem", MEM_PER_TASK, util.Role(role)),
 				},
 			}
 			log.Infof("Prepared task: %s with offer %s for launch\n", task.GetName(), offer.Id.GetValue())
--- a/mesosutil/mesosprotoutil.go
+++ b/mesosutil/mesosprotoutil.go
@@ -36,28 +36,49 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
 	return result
 }
 
-func NewScalarResource(name string, val float64) *mesos.Resource {
-	return &mesos.Resource{
+// A ResourceOpt is a functional option type for mesos.Resources.
+type ResourceOpt func(*mesos.Resource)
+
+// Role returns a ResourceOpt that sets the Role of a
+// *mesos.Resource to the given value.
+func Role(s *string) ResourceOpt {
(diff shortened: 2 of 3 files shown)


mesosutil/mesosprotoutil.go
```diff
@@ -36,27 +36,30 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
 	return result
 }
 
-func NewScalarResource(name string, val float64) *mesos.Resource {
+func NewScalarResource(name string, val float64, role *string) *mesos.Resource {
```
tsenart: Passing pointers to strings through function arguments isn't required.

mesosutil/mesosprotoutil.go
```diff
@@ -36,27 +36,30 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
...
 	return &mesos.Resource{
 		Name:   proto.String(name),
 		Type:   mesos.Value_RANGES.Enum(),
 		Ranges: &mesos.Value_Ranges{Range: ranges},
+		Role:   role,
 	}
 }
 
-func NewSetResource(name string, items []string) *mesos.Resource {
+func NewSetResource(name string, items []string, role *string) *mesos.Resource {
```
tsenart: These three functions got polluted with this argument which isn't always required. It's better to create a new function which takes a slice of resources and sets the role on them.
```go
// A ResourceOpt is a functional option type for mesos.Resources.
type ResourceOpt func(*mesos.Resource)

// NewResource returns a new *mesos.Resource with the given
// ResourceOpts applied to it.
func NewResource(opts ...ResourceOpt) *mesos.Resource {
    var r mesos.Resource
    for _, opt := range opts {
        opt(&r)
    }
    return &r
}

// Role returns a ResourceOpt that sets the Role of a
// *mesos.Resource to the given value.
func Role(s string) ResourceOpt {
    return func(r *mesos.Resource) {
        r.Role = s
    }
}
```

clintonskitson:
@tsenart check the updates and see if this is more in line with what you would like to see.

examples/example_scheduler.go
```diff
@@ -56,6 +56,7 @@ var (
 	taskCount           = flag.String("task-count", "5", "Total task count to run.")
 	mesosAuthPrincipal  = flag.String("mesos_authentication_principal", "", "Mesos authentication principal.")
 	mesosAuthSecretFile = flag.String("mesos_authentication_secret_file", "", "Mesos authentication secret file.")
+	role                = flag.String("role", "*", "Role to login framework with")
```
tsenart: A framework doesn't **login** with a role, it gets assigned a role. Here, however, you're only assigning the role to resources, not the framework.

mesosutil/mesosprotoutil.go
```diff
@@ -36,28 +36,49 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
...
 }
 
-func NewScalarResource(name string, val float64) *mesos.Resource {
-	return &mesos.Resource{
+// A ResourceOpt is a functional option type for mesos.Resources.
+type ResourceOpt func(*mesos.Resource)
+
+// Role returns a ResourceOpt that sets the Role of a
+// *mesos.Resource to the given value.
+func Role(s *string) ResourceOpt {
```
tsenart: It's not safe, necessary nor idiomatic to pass string pointers around.

mesosutil/mesosprotoutil.go
```diff
@@ -36,28 +36,49 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
...
+	return func(r *mesos.Resource) { r.Role = s }
+}
+
+func NewScalarResource(name string, val float64, resOpt ...ResourceOpt) *mesos.Resource {
```
tsenart: At this stage, it's  OK to break the API for the sake of consistency. Users of this library need to vendor it to get reproducible builds. Hence, I'd like to have a single `func NewResource(opts ...ResourceOpt) *mesos.Resource` and create the necessary `ResourceOpts` to set every resource field.

jdef:
closing this out due to old age. please re-open if/when interested.

PR #171, opened 2015-09-10
Scope Godeps, remove Makefile and adapt examples

tsenart:
Closes #148


jdef:
- dislike the removal of the Makefile, it provides useful shorthand
commands.
- also dislike the removal of Godeps without replacing it something else
(like go 1.5x vendoring, for example); they provide a reference set of 3rd
party libs that run the example code

I wonder how fun it will be to rebase the moved example code... :(

tsenart:
> - dislike the removal of the Makefile, it provides useful shorthand commands.
We have currently no need for the Makefile as the go tool covers it all for now. I'm choosing simplicity over convenience.
> also dislike the removal of Godeps without replacing it something else (like go 1.5x vendoring, for example); they provide a reference set of 3rd party libs that run the example code
AFAIU we weren't really using Godeps for anything anymore. Libraries vendoring their own dependencies is also an anti-pattern. Importers of this library are the ones to vendor this and the other libraries they depend on.

jdef:
Running the examples against a running mesos cluster are (currently) an
important piece of integration testing for mesos-go. Making this harder for
people (myself included) is very counter-productive.

tsenart:
@jdef: By no means do I want to make anyone's life harder without proper reasons. But then the vendoring should be in `examples/{scheduler,executor}` and not in the project root.

jdef:
I think I'm OK with moving the Godeps + Makefile to examples/

tsenart:
Why do you need the Makefile? You can even directly `go run` the examples...

jdef:
why does BASH have an alias command?

It's a convenience for development. It's not a huge, ugly Makefile and the
overhead of maintaining such a small file is negligible.

Also, I wonder about the consequences of moving the Godeps/ directory? I've
never seen it anywhere else but at the top-level directory for a project.

tsenart:
@jdef: BASH alias aren't usually used in a shared project, but instead in personal development environments. The same goes for a Makefile who's sole purpose is convenience. Simplicity first :-)

I also wonder if we actually need to vendor the dependencies in the examples. Wouldn't we want to know when they break with library updates?

jdef:
There's a balance to be struck here: knowing when libraries cause breakage
vs. blocking development process on bug fixes because the latest library
release breaks your integration testing.

tsenart:
Not vendoring the `examples` dependencies wouldn't block development process or bug fixes on the library itself. I think we need a better solution for integration testing.

tsenart:
As agreed on our call, we scoped `Godeps` to the examples directory.

Issue #166, opened 2015-09-09
deprecated feature parity with libmesos


spacejam:
libmesos has a grace period during which it will try to support the API from the last version, in order to support graceful cluster upgrades.

Libmesos-based mesos cluster upgrade story from mesos master version X -> Y
1. ensure that all slave+executor+scheduler processes are running on top of libmesos version X or Y
2. upgrade a mesos master to version Y, bounce it, restart the other HA masters (still running X) so that leadership passes to the master on version Y
3. observe cluster health metrics, being on high-alert to turn off the new master if things get wonky so that the old master version will be reverted to
4. after suspicion fades, upgrade and bounce the other HA masters from X to Y

mesos-go based cluster upgrade story:
1. restart everything as fast as you can, because there is no graceful transition possible, eating a guaranteed maintenance window of unavailability
2. hope in your heart of hearts that nothing bad happens while nervously staring at the health metrics - meanwhile every second you are losing money and customer trust

Furthermore, many users of mesos-go have no idea that it behaves differently from what they're used to with libmesos!  mesos-go needs to make this super explicit up-front, because outages are probably already happening due to this mismatch of assumptions!

spacejam:
This will benefit significantly from an automated test matrix, for instance the 0.23.0 branch should have basic tests that run against each of the 16 combinations ({0.23.0,0.22.0} X {master, slave, executor, scheduler}).

tsenart:
So the gist of this issue is to transition to enable a Mesos cluster upgrade story identical to libmesos, with mesos-go.

tsenart:
By @spacejam in #165
> Problem:  I'm running mesos version X.  What git SHA should I be using?
> Initially, I'd like to see a tag for:
> 0.22.0
> 0.23.0 (after #159 is merged in and more testing is performed)
> new versions after they have been properly tested

jdef:
hopefully this will be easier to do once Mesos v1 APIs land in master

PR #162, opened 2015-09-05
extracted zk client rewrite from #159

jdef:
TODO:
- [x] upgrade github.com/samuel/go-zookeeper/zk package

PR #159, opened 2015-08-17
Align uuid/ack handling with core scheduler driver, and MORE

jdef:
motivation:
- fixes #156 

other things that happened on the way:
- i want all the unit tests passing (kill the data races!)
- i want the synchronization code in the scheduler and executor to be (more) reasonable (helps with the above)
- updated godeps to latest samuel/go-zookeeper
- `make check` should ignore files in `Godeps/`
- use travis' container-based infra
- hijack the net/http server connections to implement decoding that's compat with mesos libprocess
- added support for ExitedExecutor message (--> Scheduler.ExecutorLost)

todo:
- [x] unit test for `updateRequiresAck` logic
- [x] fix remaining zk test flakes
- [x] test example scheduler/executor against running mesos cluster
- [x] the HTTP lib workarounds currently require go1.5 -- this probably won't work for all frameworks that require 1.3/1.4 API compat.

xref https://issues.apache.org/jira/browse/MESOS-2865

PR #174, opened 2015-09-11
initial support for MasterInfo as JSON

jdef:
Has not been tested against v0.24

### DIFF
--- a/detector/zoo/detect.go
+++ b/detector/zoo/detect.go
@@ -19,6 +19,7 @@
 package zoo
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"net/url"
@@ -37,6 +38,7 @@ import (
 const (
 	// prefix for nodes listed at the ZK URL path
 	nodePrefix                    = "info_"
+	nodeJsonPrefix                = "json.info_"
 	defaultMinDetectorCyclePeriod = 1 * time.Second
 )
 
@@ -49,6 +51,8 @@ type zkInterface interface {
 	data(string) ([]byte, error)
 	watchChildren(string) (string, <-chan []string, <-chan error)
 }
+
+type infoCodec func(path, node string) (*mesos.MasterInfo, error)
 
 // Detector uses ZooKeeper to detect new leading master.
 type MasterDetector struct {
@@ -121,7 +125,8 @@ func (md *MasterDetector) childrenChanged(path string, list []string, obs detect
 }
 
 func (md *MasterDetector) notifyMasterChanged(path string, list []string, obs detector.MasterChanged) {
-	topNode := selectTopNode(list)
+	// mesos v0.24 writes JSON only, v0.23 writes json and protobuf, v0.22 and prior only write protobuf
+	topNode, codec := md.selectTopNode(list)
 	if md.leaderNode == topNode {
 		log.V(2).Infof("ignoring children-changed event, leader has not changed: %v", path)
 		return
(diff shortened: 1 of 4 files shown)


jdef:
FWIW the sample detector included here (for integration testing) works much better when cherry-picked into #166 because of the bugs fixed there. I'm happy to fold these changes into that PR and re-push it as soon as I get my GH ops back.

tsenart:
OK, I'd say we wait for #162. What's your take?

detector/zoo/detect.go
```diff
@@ -37,12 +38,15 @@ import (
 const (
 	// prefix for nodes listed at the ZK URL path
 	nodePrefix                    = "info_"
+	nodeJsonPrefix                = "json.info_"
```
tsenart: s/Json/JSON/

tsenart:
LGTM

jdef:
the changes from #162 are already a part of #159 which is ready to land. so
this can go in after that and we can just close out #162

jdef:
rebased to latest master (port merge of PR159)

tsenart:
@jdef: The comment regarding JSON instead of Json can be applied everywhere else :-)

examples/zkdetect/main.go
```diff
@@ -0,0 +1,53 @@
+package main
```
tsenart: Can't we use `testing` examples instead? http://golang.org/pkg/testing/#hdr-Examples

tsenart:
Have you tested this against with 0.23.0 and 0.24.0 now?

examples/zkdetect/main.go
jdef: i want to run this on the command line. my understanding of testing
examples is that go-test runs them but we don't get separate binaries that
we can execute manually

jdef:
tested against 0.23.0 -- for some reason my masters are still outputting
only the protobuf info's; JSON parsing is not being invoked
not tested against 0.24.0

examples/zkdetect/main.go
tsenart: @jdef: Why would you want to execute a separate binary? In any case, you can easily do that by compiling the test binary and then running the example you want: `$ go test -c && ./pkg.test -test.run=ExampleXXX`

examples/Makefile
```diff
@@ -1,6 +1,9 @@
-.PHONY: all
```
tsenart: Makefiles are dead. Long live Makefiles! ;-)

examples/zkdetect/main.go
tsenart: But even without compiling the test binary: `go test -run=ExampleXXX`

tsenart:
So, assuming it's going to be too much effort to write tests for this change set at this point, can we at least do some manual testing with Mesos 0.24.0 before merging this in?

jdef:
Yes, would like to do manual testing. Need to upgrade my test cluster to
0.24

tsenart:
Great!

examples/zkdetect/main.go
jdef: if it's not easy to remember/invoke, then I (and probably most others) won't actually do it. building the zkdetect binary provides a simple entrypoint for testing and doesn't rely on a dev having to remember the specific `go test` syntax. if you really don't want the binary, then I'd suggest wrapping the go incantation in a shell script so that it's easy to invoke

jdef: the go examples also run during unit testing. the `zkdetect` bin really isn't meant for that since it relies on an external zk process/cluster to be up and running.

tsenart: > the go examples also run during unit testing. the zkdetect bin really isn't meant for that since it relies on an external zk process/cluster to be up and running.

OK. This makes sense.

tsenart: Although, we could use build tags for such. Like `+integration`. But never mind, that can be done in the future.

jdef:
tested against mesos:tip, JSON info parsing seems to work OK

PR #176, opened 2015-09-14
Fix CI

### LINKED ISSUE #169: Improve CI and fix tests
Currently all our tests are breaking. Let's fix them and get some proper CI in place.

tsenart:
This change set removes some non-determinism in the messenger package tests
which caused them to fail with port allocation conflicts.

We're removing `TestMessengerFailToStart` because testing that a kernel doesn't
allow the two sockets to bind to a single port (unless using SO_REUSEPORT in Linux)
isn't the responsibility of a user-space program.

Finally, the `mesosproto` package has been re-generated.

Fixes #169


messenger/http_transporter_test.go
```diff
@@ -289,10 +289,6 @@ func TestMutatedHostUPid(t *testing.T) {
 	if receiver.upid.Host != "127.0.0.1" {
 		t.Fatalf("reciever.upid.Host was expected to return %s, got %s\n", serverHost, receiver.upid.Host)
 	}
-
-	if receiver.upid.Port != strconv.Itoa(serverPort) {
-		t.Fatalf("receiver.upid.Port was expected to return %d, got %s\n", serverPort, receiver.upid.Port)
-	}
```
tsenart: As discussed with @jdef, the intent of this test is still valid. Find a way to test this at a different level.

jdef:
minor comment, otherwise lgtm

PR #178, opened 2015-09-15
Release v0.0.1

### LINKED ISSUE #167: 0.24.0 + (deprecated 0.23.0) support with tagged release
In the spirit of https://github.com/mesos/mesos-go/issues/166, mesos-go needs support for 0.24.0 that also retains compatibility with 0.23.0.

This means that during leader detection, it should try to deserialize the master info from ZK using both new and old formats, at least for this release.  This can be removed in the 0.25.0 (deprecated 0.24.0) release.

tsenart:
This PR introduces tagged semantic versioned releases and a `CHANGELOG`. Once the PR is merged, I'll tag the commit with two tags:

1. **v0.0.1**: mesos-go semantic version
2. **mesos-0.24.0**: mesos version compatibility

Fixes #167

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,3 @@
+2015-09-15: v0.0.1
+  First tagged release ever!
+  Mesos 0.24.0 and 0.23.0 compatibility.
--- a/mesosutil/constants.go
+++ b/mesosutil/constants.go
@@ -2,5 +2,5 @@ package mesosutil
 
 const (
 	// MesosVersion indicates the supported mesos version.
-	MesosVersion = "0.23.0"
+	MesosVersion = "0.24.0"
 )


mesosutil/constants.go
```diff
@@ -2,5 +2,5 @@ package mesosutil
 
 const (
 	// MesosVersion indicates the supported mesos version.
-	MesosVersion = "0.23.0"
```
sttts: Only one version? Then it should be called LatestSupportedMesosVersion.

CHANGELOG
```diff
@@ -0,0 +1,3 @@
+2015-09-15: v0.0.1
+  First tagged release ever!
+  Mesos 0.24.0 compatibility
```
sttts: Which versions are supported? As a baseline this info would be helpful.

mesosutil/constants.go
tsenart: No. This is the version that is supported in this git ref. If you want others, vendor with another tag.

CHANGELOG
tsenart: Thanks, I'll update the info.

mesosutil/constants.go
sttts: Which means we have json leader support, but dropped the old format?

tsenart: It means that we started supporting mesos 0.24.0. It says nothing about support for older versions which should be documented through git tags and the CHANGELOG.

sttts: This is an odd interface.

sttts: MaxMesosVersion would make sense. But mutil.MesosVersion to be potentially a different version than the installed Mesos master is strange.

tsenart:
@sttts: Made it clear that we support *also* support Mesos 0.23.0 in the CHANGELOG. I'll also find the place in the git history where we started supporting Mesos 0.23.0 and tag it accordingly.

CHANGELOG
```diff
@@ -0,0 +1,3 @@
+2015-09-15: v0.0.1
+  First tagged release ever!
+  Mesos 0.24.0 and 0.23.0 compatibility.
```
sttts: better.

tsenart:
To make it clear: Documentation of Mesos version support deprecation shall be done via the CHANGELOG.

mesosutil/constants.go
sttts: but this discussion is out of scope of this PR:

sttts:
lgtm

Issue #182, opened 2015-09-19
High CPU utilization of Executor in Example


oxtoacart:
This is the same issue as #55.  I was able to reproduce on OS X 10.10.

oxtoacart:
I've submitted a PR with a fix - #183

PR #183, opened 2015-09-19
Exiting loop on closed channel, fixes mesos/mesos-go#182

oxtoacart:


examples/executor/main.go
```diff
@@ -64,6 +65,8 @@ func (exec *exampleExecutor) LaunchTask(driver exec.ExecutorDriver, taskInfo *me
 	// this is where one would perform the requested task
 	//
 
+	time.Sleep(5 * time.Second)
```
oxtoacart: Added this to keep executor running long enough to see problem.

messenger/http_transporter.go
```diff
@@ -490,7 +490,10 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
 	// libprocess will ignore responses, see ignore_data).
 	// see https://github.com/apache/mesos/blob/adecbfa6a216815bd7dc7d26e721c4c87e465c30/3rdparty/libprocess/src/process.cpp#L2192
 	for {
-		if r, ok := <-decoder.Requests(); ok {
+		if r, ok := <-decoder.Requests(); !ok {
```
oxtoacart: This was the problem - decoder.Requests() closes, this turns into a tight loop that just keeps trying to read from the already closed channel.  The solution is to exit the loop once the channel is closed.

examples/executor/main.go
jdef: understood; needs removal prior to merge

messenger/http_transporter.go
jdef: nice catch. when there's a 'return' as a result of an if, we're trying to stick with a style like this going forward:
```go
if something {
  return
}
// do something else
```
this means declaring the r, ok outside of the if statement, and that's OK

oxtoacart: Done.

examples/executor/main.go
oxtoacart: Done.

messenger/http_transporter.go
```diff
@@ -490,52 +490,55 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
...
-				log.V(2).Infof("Receiving %s %v from %v, length %v\n", r.Method, r.URL, from, len(data))
-				m := &Message{
-					UPID:  from,
-					Name:  extractNameFromRequestURI(r.RequestURI),
-					Bytes: data,
-				}
+		r, ok := <-decoder.Requests()
+		if !ok {
+			// Channel closed, return
+			return
```
jdef: I missed this in my earlier review... should probably `break` here instead of `return` so that if there's a waiting error in the decoder it'll be logged.

jdef:
minor comment remaining to be addressed. once fixed, and commits sqashed, lgtm

oxtoacart:
Addressed comment, squashed commits into one

jdef:
thanks. LGTM

jdef:
fixes #182

PR #186, opened 2015-09-22
Fixing default path to example executor

drewkerrigan:
The examples Makefile now creates a binary called "executor" rather than "example_executor", the default parameter value for the "executor" flag in the example scheduler should be updated.


tsenart:
LGTM

PR #189, opened 2015-09-22
fix racy UPID access and intermittent scheduler deadlock at shutdown time

jdef:
- UPID pointer sharing between transporter and messenger was awful - cleaned it up
- during integration testing, noticed that scheduler sometimes hung at shutdown - fixed that
  - the `withScheduler` construct is looking uglier and uglier, though it is somewhat contained to the factory func that builds the func; there must be a simpler solution here.
- libprocess decoder fixes
  - check for ECONNRESET for graceful connection termination
  - reading Request's should't block if shouldQuit chan is closed
- unit tests were not happy after the above changes, fixed them

### DIFF
--- a/messenger/decoder.go
+++ b/messenger/decoder.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	log "github.com/golang/glog"
@@ -523,6 +524,19 @@ func readBodyState(d *httpDecoder) httpState {
 	return readBodyState
 }
 
+func isGracefulTermSignal(err error) bool {
+	if err == io.EOF {
+		return true
+	}
+	if operr, ok := err.(*net.OpError); ok {
+		if operr.Op != "read" {
+			return false
+		}
+		return err == syscall.ECONNRESET
+	}
+	return false
+}
+
 func awaitRequestState(d *httpDecoder) httpState {
 	log.V(2).Infoln(d.idtag + "await-request-state")
 	tr := textproto.NewReader(d.rw.Reader)
@@ -530,7 +544,7 @@ func awaitRequestState(d *httpDecoder) httpState {
 		return terminateState
 	}
 	requestLine, err := tr.ReadLine()
-	if requestLine == "" && err == io.EOF {
+	if requestLine == "" && isGracefulTermSignal(err) {
 		// we're actually expecting this at some point, so don't react poorly
 		return gracefulTerminateState
 	}
--- a/messenger/decoder.go
+++ b/messenger/decoder.go
@@ -529,10 +529,7 @@ func isGracefulTermSignal(err error) bool {
 		return true
 	}
 	if operr, ok := err.(*net.OpError); ok {
-		if operr.Op != "read" {
-			return false
-		}
-		return err == syscall.ECONNRESET
+		return operr.Op == "read" && err == syscall.ECONNRESET
 	}
 	return false
 }
--- a/messenger/http_transporter.go
+++ b/messenger/http_transporter.go
@@ -424,30 +424,30 @@ func (t *HTTPTransporter) start() (upid.UPID, <-chan error) {
 	if err := t.listen(); err != nil {
 		ch <- err
 		return upid.UPID{}, ch
-	} else {
-		// TODO(yifan): Set read/write deadline.
-		go func() {
-			s := &http.Server{
-				ReadTimeout:  DefaultReadTimeout,
-				WriteTimeout: DefaultWriteTimeout,
-				Handler:      t.mux,
-			}
-			err := s.Serve(t.listener)
-			select {
-			case <-t.shouldQuit:
-				log.V(1).Infof("HTTP server stopped because of shutdown")
-				ch <- nil
-			default:
-				if err != nil && log.V(1) {
-					log.Errorln("HTTP server stopped with error", err.Error())
-				} else {
-					log.V(1).Infof("HTTP server stopped")
-				}
-				ch <- err
-				t.Stop(false)
-			}
-		}()
 	}
+
+	// TODO(yifan): Set read/write deadline.
+	go func() {
+		s := &http.Server{
+			ReadTimeout:  DefaultReadTimeout,
+			WriteTimeout: DefaultWriteTimeout,
+			Handler:      t.mux,
+		}
+		err := s.Serve(t.listener)
+		select {
+		case <-t.shouldQuit:
+			log.V(1).Infof("HTTP server stopped because of shutdown")
+			ch <- nil
+		default:
+			if err != nil && log.V(1) {
+				log.Errorln("HTTP server stopped with error", err.Error())
+			} else {
+				log.V(1).Infof("HTTP server stopped")
+			}
+			ch <- err
+			t.Stop(false)
+		}
+	}()
 	return t.upid, ch
 }
 
@@ -504,11 +504,7 @@ func (t *HTTPTransporter) processRequests(from *upid.UPID, incoming <-chan *Requ
 	for {
 		select {
 		case r, ok := <-incoming:
-			if !ok {
-				// Channel closed, break out of loop
-				return
-			}
-			if !t.processOneRequest(from, r) {
+			if !ok || !t.processOneRequest(from, r) {
 				return
 			}
 		case <-t.shouldQuit:
--- a/messenger/http_transporter.go
+++ b/messenger/http_transporter.go
@@ -488,67 +488,7 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
 	decoder := DecodeHTTP(w, r)
 	defer decoder.Cancel(true)
 
-	//TODO(jdef) Only send back an HTTP response if this isn't from libprocess
-	// (which we determine by looking at the User-Agent). This is
-	// necessary because older versions of libprocess would try and
-	// recv the data and parse it as an HTTP request which would
-	// fail thus causing the socket to get closed (but now
-	// libprocess will ignore responses, see ignore_data).
-	// see https://github.com/apache/mesos/blob/adecbfa6a216815bd7dc7d26e721c4c87e465c30/3rdparty/libprocess/src/process.cpp#L2192
-rcvLoop:
-	for {
-		var request *Request
-		select {
-		case r, ok := <-decoder.Requests():
-			if !ok {
-				// Channel closed, break out of loop
-				break rcvLoop
-			}
-			request = r
-		case <-t.shouldQuit:
-			break rcvLoop
-		}
-		if func() bool {
-			// regardless of whether we write a Response we must close this chan
-			defer close(request.response)
-
-			//TODO(jdef) this is probably inefficient given the current implementation of the
-			// decoder: no need to make another copy of data that's already competely buffered
-			data, err := ioutil.ReadAll(request.Body)
-			if err != nil {
-				// this is unlikely given the current implementation of the decoder:
-				// the body has been completely buffered in memory already
-				log.Errorf("failed to read HTTP body: %v\n", err)
-				return true
-			}
-			log.V(2).Infof("Receiving %s %v from %v, length %v\n", request.Method, request.URL, from, len(data))
-			m := &Message{
-				UPID:  from,
-				Name:  extractNameFromRequestURI(request.RequestURI),
-				Bytes: data,
-			}
-
-			keepGoing := true
-			select {
-			case <-t.shouldQuit:
-				keepGoing = false
-			case t.messageQueue <- m:
-			}
-
-			// if user-agent != libprocess then we need to send a response
-			if _, ok := parseLibprocessAgent(request.Header); !ok {
-				log.V(2).Infof("not libprocess agent, sending a 202")
-				request.response <- Response{
-					code:   202,
-					reason: "Accepted",
-				}
-			}
-
-			return keepGoing
-		}() {
-			continue
-		}
-	}
+	t.processRequests(from, decoder.Requests())
 
 	// log an error if there's one waiting, otherwise move on
 	select {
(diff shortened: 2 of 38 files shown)


jdef:
rebased to origin/master

scheduler/scheduler.go
```diff
@@ -246,6 +236,86 @@ func NewMesosSchedulerDriver(config DriverConfig) (initializedDriver *MesosSched
...
+				}
+			default:
+			}
+			return
+		}
+		for !trySend() {
+			t.Reset(timeout) // TODO(jdef) add jitter to this
+		}
+		// have to do this outside trySend because here we're guarded by eventLock; it's ok
+		// if this happens more then once.
```
jdef: s/then/than/

scheduler/scheduler.go
```diff
@@ -755,13 +812,14 @@ func (driver *MesosSchedulerDriver) start() (mesos.Status, error) {
...
 	})
 
-	// register with Detect() AFTER we have a self pid from the messenger, otherwise things get ugly
-	// because our internal messaging depends on it. detector callbacks are routed over the messenger
-	// bus, maintaining serial (concurrency-safe) callback execution.
-	log.V(1).Infof("starting master detector %T: %+v", driver.masterDetector, driver.masterDetector)
-	driver.masterDetector.Detect(listener)
-
-	log.V(2).Infoln("master detector started")
+	if driver.masterDetector != nil {
```
jdef: making masterDetector optional keeps setup for unit tests simpler, but i'm not sure that I like the checks for nil in this class. i'd almost rather have a NoopDetector implementation that unit tests used (guaranteed to do nothing all the time).

scheduler/scheduler.go
```diff
@@ -887,18 +948,31 @@ func (driver *MesosSchedulerDriver) Join() (mesos.Status, error) {
...
-	if stat := driver.status; stat != mesos.Status_DRIVER_RUNNING {
-		return stat, fmt.Errorf("Unable to Join, expecting driver status %s, but is %s", mesos.Status_DRIVER_RUNNING, stat)
+func (driver *MesosSchedulerDriver) join() (stat mesos.Status, err error) {
+	if stat = driver.status; stat != mesos.Status_DRIVER_RUNNING {
+		err = fmt.Errorf("Unable to Join, expecting driver status %s, but is %s", mesos.Status_DRIVER_RUNNING, stat)
+		return
 	}
+
+	timeout := 1 * time.Second
+	t := time.NewTimer(timeout)
```
jdef: could probably use a ticker here instead

scheduler/scheduler_unit_test.go
```diff
@@ -354,30 +305,30 @@ func (suite *SchedulerTestSuite) TestSchdulerDriverStop_WithoutFailover() {
...
 	// Set expections and return values.
-	messenger := &msgTracker{MockedMessenger: mockedMessenger()}
-	driver := newTestSchedulerDriver(suite.T(), driverConfigMessenger(NewMockScheduler(), suite.framework, suite.master, nil, messenger))
+	mess := &msgTracker{MockedMessenger: mockedMessenger()}
+	d := DriverConfig{
+		Scheduler:    NewMockScheduler(),
+		Framework:    suite.framework,
+		Master:       suite.master,
+		NewMessenger: func() (messenger.Messenger, error) { return mess, nil },
+		NewDetector:  func() (detector.Master, error) { return nil, nil },
```
jdef: here's where we could use a NoopDetector

jdef:
drawing a line in the sand here (with respect to refactoring, etc). i've met the original objective of this PR (and then some) re: fixing the UPID pointer sharing between messenger and transport.

@tsenart PTAL

messenger/decoder.go
```diff
@@ -523,6 +524,19 @@ func readBodyState(d *httpDecoder) httpState {
...
 
+func isGracefulTermSignal(err error) bool {
+	if err == io.EOF {
+		return true
+	}
+	if operr, ok := err.(*net.OpError); ok {
+		if operr.Op != "read" {
+			return false
+		}
+		return err == syscall.ECONNRESET
```
tsenart: `return operr.Op == "read" && operr.Err == syscall.ECONNRESET`

tsenart: Or even:
```go
operr, ok := err.(*net.OpError)
return ok && operr.Op == "read" && operr.Err == syscall.ECONNRESET
```

messenger/decoder.go
```diff
@@ -523,6 +524,19 @@ func readBodyState(d *httpDecoder) httpState {
 	return readBodyState
 }
 
+func isGracefulTermSignal(err error) bool {
```
tsenart: Is this name appropriate? AFAIK `ECONNRESET` doesn't mean the connection was *gracefully* closed on the other side, only that it was closed.

messenger/http_transporter.go
```diff
@@ -85,10 +85,10 @@ func (s *stoppedState) Recv() (*Message, error)                        { return
 func (s *stoppedState) Inject(ctx context.Context, msg *Message) error { return errTerminal }
 func (s *stoppedState) Stop(graceful bool) error                       { return errTerminal }
 func (s *stoppedState) Install(messageName string)                     {}
-func (s *stoppedState) Start() <-chan error {
+func (s *stoppedState) Start() (upid.UPID, <-chan error) {
 	ch := make(chan error, 1)
 	ch <- errTerminal
-	return ch
+	return upid.UPID{}, ch
```
tsenart: I'm not familiar with this code. Why are we returning an empty `UPID` here and the other implementations of this interface below?

messenger/http_transporter.go
```diff
@@ -412,17 +412,18 @@ func (t *HTTPTransporter) listen() error {
...
 }
 
 // start expects to be guarded by stateLock
-func (t *HTTPTransporter) start() <-chan error {
+func (t *HTTPTransporter) start() (upid.UPID, <-chan error) {
 	ch := make(chan error, 1)
 	if err := t.listen(); err != nil {
 		ch <- err
+		return upid.UPID{}, ch
 	} else {
```
tsenart: We can get rid of the else branch due to the early return.

messenger/http_transporter.go
```diff
@@ -482,65 +488,87 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
...
-			if err != nil {
-				// this is unlikely given the current implementation of the decoder:
-				// the body has been completely buffered in memory already
-				log.Errorf("failed to read HTTP body: %v\n", err)
-				return true
+func (t *HTTPTransporter) processRequests(from *upid.UPID, incoming <-chan *Request) {
+	for {
+		select {
+		case r, ok := <-incoming:
+			if !ok {
```
tsenart:
```go
if !ok || !t.processOneRequest(from, r) {
    return
```

messenger/http_transporter.go
```diff
@@ -482,65 +488,87 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
...
 
-			return keepGoing
-		}() {
-			continue
+	// deterministic behavior and output..
+	select {
+	case <-t.shouldQuit:
+		keepGoing = false
+		select {
+		case t.messageQueue <- m:
```
tsenart: Best effort?

messenger/http_transporter.go
```diff
@@ -482,65 +488,87 @@ func (t *HTTPTransporter) messageDecoder(w http.ResponseWriter, r *http.Request)
...
+		default:
+		}
+	case t.messageQueue <- m:
+		select {
+		case <-t.shouldQuit:
```
tsenart: Since we're already receiving from `t.shouldQuit` in `processRequests`, why do we have to do it here too?

messenger/http_transporter_test.go
```diff
@@ -17,11 +19,8 @@ import (
 )
 
 func TestTransporterNew(t *testing.T) {
-	id, err := upid.Parse(fmt.Sprintf("mesos1@localhost:%d", getNewPort()))
-	assert.NoError(t, err)
-	trans := NewHTTPTransporter(id, nil)
+	trans := NewHTTPTransporter(upid.UPID{ID: "mesos1", Host: "localhost"}, nil)
```
tsenart: Why aren't we passing the `id` in like we did before?

messenger/http_transporter_test.go
```diff
@@ -161,23 +151,20 @@ func TestTransporterStartAndRcvd(t *testing.T) {
...
 
 	// setup sender (client) process
-	sndUpid, err := upid.Parse(fmt.Sprintf("mesos1@localhost:%d", getNewPort()))
-	assert.NoError(t, err)
-
-	sender := NewHTTPTransporter(sndUpid, nil)
+	sender := NewHTTPTransporter(upid.UPID{ID: "mesos1", Host: "localhost"}, nil)
 	msg := &Message{
-		UPID:         rcvPid,
+		UPID:         &rcvPid,
```
tsenart: Should we still use UPID pointers in `Message` structs?

messenger/http_transporter_test.go
```diff
@@ -337,3 +302,49 @@ func makeMockServer(path string, handler func(rsp http.ResponseWriter, req *http
...
+		shouldQuit:   make(chan struct{}),
+	}
+	testfunc := func(expectProceed bool) {
+		rchan := make(chan Response, 1)
+		proceed := ht.processOneRequest(&upid.UPID{ID: "james"}, &Request{
+			response: rchan,
+			Request: &http.Request{
+				Method:     "foo",
+				RequestURI: "a/z/bar",
+				Body:       ioutil.NopCloser(bytes.NewBuffer([]byte{})),
```
tsenart: `s/bytes.NewBuffer/bytes.Reader{}`

messenger/messenger.go
```diff
@@ -214,8 +198,10 @@ func (m *MesosMessenger) Send(ctx context.Context, upid *upid.UPID, msg proto.Me
 // 1) routing internal error to callback handlers
 // 2) testing components without starting remote servers.
 func (m *MesosMessenger) Route(ctx context.Context, upid *upid.UPID, msg proto.Message) error {
-	// if destination is not self, send to outbound.
-	if !upid.Equal(m.upid) {
+	if upid == nil {
+		panic("cannot route a message to a nil pid")
```
tsenart: Why don't we change the signature of the method to not take a pointer instead?

messenger/messenger.go
```diff
@@ -229,30 +215,29 @@ func (m *MesosMessenger) Route(ctx context.Context, upid *upid.UPID, msg proto.M
 	return m.tr.Inject(ctx, &Message{upid, name, msg, data})
 }
 
-// Start starts the messenger.
+// Start starts the messenger; expects to be called once and only once.
```
tsenart: https://golang.org/pkg/sync/#Once ?

messenger/messenger.go
```diff
@@ -37,21 +35,8 @@ import (
...
 )
 
-var (
-	sendRoutines   int
-	encodeRoutines int
-	decodeRoutines int
-)
-
-func init() {
-	flag.IntVar(&sendRoutines, "send-routines", 1, "Number of network sending routines")
```
tsenart: :+1:

scheduler/scheduler.go
```diff
@@ -88,6 +100,7 @@ type DriverConfig struct {
 	BindingPort      uint16                                // optional
 	PublishedAddress net.IP                                // optional
 	NewMessenger     func() (messenger.Messenger, error)   // optional
+	NewDetector      func() (detector.Master, error)       // optional
```
tsenart: Why don't we pass a `Messenger` and `Master` directly in the `DriverConfig`?

scheduler/scheduler.go
```diff
@@ -246,6 +236,86 @@ func NewMesosSchedulerDriver(config DriverConfig) (initializedDriver *MesosSched
 	return
 }
 
+func (driver *MesosSchedulerDriver) makeWithScheduler(cs Scheduler) func(func(Scheduler)) {
```
tsenart: OMG, I think I need a vacation after before reading this function. :-)

tsenart:
I can't entirely say it LGTM, but it definitely SGTM :-)

tsenart:
... given we're gonna start from scratch anyways.

messenger/decoder.go
jdef: That's about as graceful as libprocess gets

messenger/http_transporter.go
jdef: We don't want to return a valid UPID in the case of an error condition.

messenger/http_transporter.go
jdef: that, and deterministic behavior. validated by unit test.

messenger/http_transporter.go
jdef: deterministic behavior

messenger/http_transporter_test.go
jdef: Parsing the UPID was a lazy approach to UPID creation. There's no
compelling reason to do it that way.

messenger/http_transporter_test.go
jdef: Probably not, but that can be a separate refactoring.

messenger/messenger.go
jdef: Because I had to draw the line somewhere. There's still lots of places in
the code base where UPID pointers are passed around. The initial goal was
to eliminate the sharing between the transport and messenger in the case
where the "self" UPID could change in an unguarded (non-critical section)
context.

messenger/messenger.go
jdef: This code is ripe for refactoring. Not in this PR.

scheduler/scheduler.go
jdef: Me too, and I wrote it. I tried to tease things apart for readability but
it's still too complex for my taste. But it *does* solve the hanging
problems that I observed at scheduler shutdown. I can clean up the deferred
lock management so that it's less heady.

tsenart: Yeah, I totally understand. Let's not put too much brain cycles here. They'll be better spent on the new bindings.

scheduler/scheduler.go
jdef: I thought I had a reason the first time around when I wrote NewMessenger. Followed the same pattern here without much forethought. I'm open to refactoring this pattern, but please in a separate PR.

PR #198, opened 2015-10-15
be more explicit about timeouts and connection closure

jdef:
* avoid possible racy reference to `d.con`, only close it if it was ever valid
* flushing response buffers should eventually time out
* specify read timeout for initial http client
* be explicit about timeouts for dialed connections

### DIFF
--- a/messenger/decoder.go
+++ b/messenger/decoder.go
@@ -22,6 +22,12 @@ import (
 const (
 	DefaultReadTimeout  = 5 * time.Second
 	DefaultWriteTimeout = 5 * time.Second
+
+	// writeFlushPeriod is the amount of time we're willing to wait for a single
+	// response buffer to be fully written to the underlying TCP connection; after
+	// this amount of time the remaining bytes of the response are discarded. see
+	// responseWriter().
+	writeFlushPeriod = 30 * time.Second
 )
 
 type decoderID int32
@@ -75,9 +81,9 @@ type httpDecoder struct {
 	cancelGuard  sync.Mutex
 	readTimeout  time.Duration
 	writeTimeout time.Duration
-	idtag        string          // useful for debugging
-	sendError    func(err error) // abstraction for error handling
-	outCh        chan *bytes.Buffer
+	idtag        string             // useful for debugging
+	sendError    func(err error)    // abstraction for error handling
+	outCh        chan *bytes.Buffer // chan of responses to be written to the connection
 }
 
 // DecodeHTTP hijacks an HTTP server connection and generates mesos libprocess HTTP
@@ -135,27 +141,13 @@ func (d *httpDecoder) Cancel(graceful bool) {
 
 func (d *httpDecoder) run(res http.ResponseWriter) {
 	defer func() {
-		close(d.outCh)
+		close(d.outCh) // we're finished generating response objects
 		log.V(2).Infoln(d.idtag + "run: terminating")
 	}()
 
-	go func() {
-		for buf := range d.outCh {
-			select {
-			case <-d.forceQuit:
-				return
-			default:
-			}
-			//TODO(jdef) I worry about this busy-looping
-			for buf.Len() > 0 {
-				d.tryFlushResponse(buf)
-			}
-		}
-	}()
-
-	var next httpState
-	for state := d.bootstrapState(res); state != nil; state = next {
-		next = state(d)
+	for state := d.bootstrapState(res); state != nil; {
+		next := state(d)
+		state = next
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jdef:
@sttts PTAL

messenger/decoder.go
```diff
@@ -139,22 +139,36 @@ func (d *httpDecoder) run(res http.ResponseWriter) {
...
 		for buf := range d.outCh {
-			select {
-			case <-d.forceQuit:
-				return
-			default:
-			}
 			//TODO(jdef) I worry about this busy-looping
-			for buf.Len() > 0 {
+			now := time.Now()
+			for buf.Len() > 0 && time.Since(now) < 30*time.Second {
```
karlkfi: the logic in this double for loop and select statement is a bit complex. Can you leave some comments about what you're doing to make it easier to read?

messenger/decoder.go
```diff
@@ -139,22 +139,36 @@ func (d *httpDecoder) run(res http.ResponseWriter) {
 		log.V(2).Infoln(d.idtag + "run: terminating")
 	}()
 
+	connCh := make(chan net.Conn, 1)
 	go func() {
+		con, ok := <-connCh
```
karlkfi: i like to use "open" as the 2nd variable name on channel returns, makes the if statement read better

messenger/http_transporter.go
```diff
@@ -44,6 +44,21 @@ var (
 	errNotStarted      = errors.New("HTTP transport has not been started")
 	errTerminal        = errors.New("HTTP transport is terminated")
 	errAlreadyRunning  = errors.New("HTTP transport is already running")
+
+	httpTransport, httpClient = func() (tr *http.Transport, cl *http.Client) {
```
karlkfi: why is this done in a function instead of just in a var definition?

messenger/http_transporter.go
```diff
@@ -124,13 +139,12 @@ type HTTPTransporter struct {
...
 func NewHTTPTransporter(upid upid.UPID, address net.IP) *HTTPTransporter {
-	tr := &http.Transport{}
 	result := &HTTPTransporter{
 		upid:         upid,
 		messageQueue: make(chan *Message, defaultQueueSize),
 		mux:          http.NewServeMux(),
-		client:       &http.Client{Transport: tr},
-		tr:           tr,
+		client:       httpClient,
+		tr:           httpTransport,
```
karlkfi: Why is the transport supplied to the HTTPTransporter if you can just pull it out of the Client with `t.client.Transport`?

jdef: You'd have to cast it because http.Client.Transport is a RoundTripper
interface. I'd rather not cast.

messenger/http_transporter.go
jdef: Because I'm assigning both vars at the same time, client depends on
transport. Given that assignments are guaranteed to be carried out in
left-to-right order (just learned something new) I guess I don't need the
func.

sttts:
About the commit msg: the description of the PR belongs into the commit msg.

messenger/decoder.go
```diff
@@ -22,6 +22,7 @@ import (
 const (
 	DefaultReadTimeout  = 5 * time.Second
 	DefaultWriteTimeout = 5 * time.Second
+	WriteFlushPeriod    = 30 * time.Second
```
sttts: Please comment.

jdef:
addressed review comments; squashed and commit msg updated

messenger/decoder.go
```diff
@@ -373,11 +366,44 @@ func (d *httpDecoder) bootstrapState(res http.ResponseWriter) httpState {
 		d.sendError(errHijackFailed)
 		return terminateState
 	}
+
 	d.rw = rw
 	d.con = c
+
+	go d.responseWriter()
```
sttts: Much much much better :)

sttts:
lgtm.

Is there a possibility to test the new behavior?

jdef:
tomas and I haven't been super-concerned about tons of unit tests for this
code. it's all being thrown away in roughly 6mo anyway once the new HTTP
API-based bindings are in place. I've run some integration tests locally on
my laptop to confirm that I probably haven't introduced any new bugs.

messenger/http_transporter.go
```diff
@@ -44,6 +44,18 @@ var (
 	errNotStarted      = errors.New("HTTP transport has not been started")
 	errTerminal        = errors.New("HTTP transport is terminated")
 	errAlreadyRunning  = errors.New("HTTP transport is already running")
+
+	httpTransport, httpClient = &http.Transport{
```
karlkfi: Are you positive this doesn't work as two declarations? I could sworn I'd done that before for this exact reason.

jdef: it might. i did not test.

karlkfi: Indeed, it does work: https://play.golang.org/p/IIvFMrxYmf

jdef: :+1:

Issue #187, opened 2015-09-22
Implement AcceptOffers in Scheduler, Add Support for Dynamic Reservations / Persistent Volumes


drewkerrigan:
The protobuf messages for 0.24 have been added, but functionality is not yet implemented to utilize them.

Here is the documentation for persistent volumes: [http://mesos.apache.org/documentation/latest/persistent-volume/](http://mesos.apache.org/documentation/latest/persistent-volume/). From the docs: "`Offer::Operation::Create` and `Offer::Operation::Destroy` messages are available for frameworks to send back via the acceptOffers API as a response to a resource offer."

The `Create` and `Destroy` messages exist, but there is currently no way to call the `acceptOffers` function in the scheduler driver directly. 

Additionally, the use of persistent volumes requires that a resource be reserved using dynamic or static reservations. Support for dynamic reservations should also be added as a prerequisite to persistent volumes in the Go binding ([http://mesos.apache.org/documentation/latest/reservation/](http://mesos.apache.org/documentation/latest/reservation/)). From the docs: "`Offer::Operation::Reserve` and `Offer::Operation::Unreserve` messages are available for frameworks to send back via the acceptOffers API as a response to a resource offer."

Similarly, these messages exist, but cannot be used until support for direct incantation of the `acceptOffers` function within the scheduler driver is implemented.

Lastly, an example of using dynamic reservations and persistent volumes should be added to the `examples/` directory. For reference, here is the only existing example scheduler that I am aware of, it is written in C++: [https://github.com/apache/mesos/blob/master/src/examples/persistent_volume_framework.cpp](https://github.com/apache/mesos/blob/master/src/examples/persistent_volume_framework.cpp).

drewkerrigan:
TL;DR: 
`scheduler.go` needs an `AcceptOffers` function or perhaps abstraction functions to `ReserveResrouce` / `UnreserveResource` (using `Offer::Operation::Reserve` and `Offer::Operation::Unreserve` messages) as well as `CreatePersistentVolume` / `DestroyPersistentVolume` (using `Offer::Operation::Create` and `Offer::Operation::Destroy` messages).

tsenart:
@drewkerrigan: Thanks for filing. Please take a look at this: https://github.com/mesos/mesos-go/pull/188

drewkerrigan:
[...]

PR #206, opened 2015-11-03
Add AcceptOffers Support

drewkerrigan:
Addressing #187 

Added `AcceptOffers` to the scheduler driver which utilizes the new HTTP API to send a Scheduler Call Message to the Mesos Master. Supported operation types are `LAUNCH`, `RESERVE`, `UNRESERVE`, `CREATE`, `DESTROY`.

I understand that work is currently happening in the [next](https://github.com/mesos/mesos-go/tree/next) branch, but until that is done, this is a way to utilize dynamic reservations / persistent volumes in the existing mesos-go binding.


messenger/messenger.go
```diff
@@ -391,5 +391,15 @@ func (m *MesosMessenger) decodeLoop() {
 
 // getMessageName returns the name of the message in the mesos manner.
 func getMessageName(msg proto.Message) string {
-	return fmt.Sprintf("%v.%v", "mesos.internal", reflect.TypeOf(msg).Elem().Name())
+	msgType := reflect.TypeOf(msg).Elem().Name()
+	var msgName string
+
+	switch msgType {
```
jdef: i'd rather see a type-switch vs. reflection:
```go
switch msg := msg.(type); msg {
case *mesos.Call:
  // ... msg is now a *mesos.Call instance
default:
  // msg could be anything else
}
```

messenger/message.go
```diff
@@ -36,7 +36,29 @@ type Message struct {
...
+	case "external":
+		requestURI = fmt.Sprintf("/api/v1/%s", m.Name)
+	default:
+		requestURI = fmt.Sprintf("/%s/%s", m.UPID.ID, m.Name)
+	}
+
+	return requestURI
+}
+
+func (m *Message) GetType() string {
```
jdef: would rather see:
```go
func (m *Message) isV1API() bool {
...
}
```

scheduler/scheduler.go
```diff
@@ -1091,6 +1091,110 @@ func (driver *MesosSchedulerDriver) abort(cause error) (stat mesos.Status, err e
...
+
+	okOperations := make([]*mesos.Offer_Operation, 0, len(operations))
+
+	for _, offerId := range offerIds {
+		for _, operation := range operations {
+			// Keep only the slave PIDs where we run tasks so we can send
+			// framework messages directly.
+			if driver.cache.containsOffer(offerId) {
+				// Validate
+				switch *operation.Type {
```
jdef: s/*operation.Type/operation.GetType()/

scheduler/scheduler.go
```diff
@@ -1091,6 +1091,110 @@ func (driver *MesosSchedulerDriver) abort(cause error) (stat mesos.Status, err e
...
+			} else {
+				log.Warningf("Attempting to accept offers with unknown offer %s\n", offerId.GetValue())
+			}
+		}
+
+		driver.cache.removeOffer(offerId) // if offer
+	}
+
+	// Accept Offers
+	accept := &mesos.Call_Accept{
```
neilconway: I think it is a bit better to construct this in a single literal:

```
message := &mesos.Call{
    FrameworkId: ...,
    Type: mesos.Call_ACCEPT.Enum(),
    Accept: &mesos.Call_Accept{
        ...
    },
}```

File: scheduler/scheduler.go [hunk 5]
```diff
@@ -1091,6 +1091,110 @@ func (driver *MesosSchedulerDriver) abort(cause error) (stat mesos.Status, err e
...
+		return driver.status, fmt.Errorf("Not connected to master.")
+	}
+
+	okOperations := make([]*mesos.Offer_Operation, 0, len(operations))
+
+	for _, offerId := range offerIds {
+		for _, operation := range operations {
+			// Keep only the slave PIDs where we run tasks so we can send
+			// framework messages directly.
+			if driver.cache.containsOffer(offerId) {
```
neilconway: I think it is a bit clearer to invert the conditional:

```
if !driver.cache.containsOffer(offerId) {
    // warning
    continue
}

// Validate
```

drewkerrigan:
@neilconway @jdef Good feedback, I've addressed your comments in the latest commit.

scheduler/scheduler.go
```diff
@@ -1091,6 +1091,109 @@ func (driver *MesosSchedulerDriver) abort(cause error) (stat mesos.Status, err e
...
+
+		driver.cache.removeOffer(offerId) // if offer
+	}
+
+	// Accept Offers
+	message := &mesos.Call{
+		FrameworkId: driver.frameworkInfo.Id,
+		Type:        mesos.Call_ACCEPT.Enum(),
+		Accept: &mesos.Call_Accept{
+			OfferIds:   offerIds,
```
jdef: It's possible that `offerIds` will contain ID's that are not actually used (because of the earlier `continue`)

jdef: nevermind. misread code

scheduler/scheduler.go
```diff
@@ -1091,6 +1091,109 @@ func (driver *MesosSchedulerDriver) abort(cause error) (stat mesos.Status, err e
...
 
+func (driver *MesosSchedulerDriver) AcceptOffers(offerIds []*mesos.OfferID, operations []*mesos.Offer_Operation, filters *mesos.Filters) (mesos.Status, error) {
+	driver.eventLock.Lock()
+	defer driver.eventLock.Unlock()
+
+	if stat := driver.status; stat != mesos.Status_DRIVER_RUNNING {
+		return stat, fmt.Errorf("Unable to AcceptOffers, expected driver status %s, but got %s", mesos.Status_DRIVER_RUNNING, stat)
+	}
+
+	if !driver.connected {
```
jdef: actually you probably want to send TASK_LOST to the Scheduler for each LAUNCH op. see launchTasks() and `pushLostTask`. the native code does this: https://github.com/apache/mesos/blob/master/src/sched/sched.cpp#L1160

drewkerrigan: Ok, this is added

drewkerrigan:
Ah yeah, good catch, adding

drewkerrigan:
ping @jdef @neilconway, I added some unit tests as well as an example scheduler that uses the new functionality. Let me know if you need additional changes, or if anything else is preventing a merge.

jdef:
lgtm :ship:

PR #208, opened 2015-11-10
Add missing methods to the MockScheduler

alde:


jdef:
instead of adding these funcs to `MockScheduler` we need a `MockSchedulerDriver` instead. Like this: https://github.com/kubernetes/kubernetes/blob/master/contrib/mesos/pkg/scheduler/components/framework/driver_mock.go

PR #209, opened 2015-11-13
next: decrease surface area between httpcli and http.Client...

jdef:
...and abstract the HTTP round-trip function

### DIFF
--- a/httpcli/http.go
+++ b/httpcli/http.go
@@ -72,6 +72,9 @@ func New(opts ...Opt) *Client { return new(Client).With(opts...) }
 // Opt defines a functional option for the HTTP client type.
 type Opt func(*Client)
 
+// RequestOpt defines a functional option for an http.Request.
+type RequestOpt func(*http.Request)
+
 // With applies the given Opts to a Client and returns itself.
 func (c *Client) With(opts ...Opt) *Client {
 	for _, opt := range opts {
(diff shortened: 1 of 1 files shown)


tsenart:
Why not use the RoundTripper interface? That was the purpose of it in the
first place.

httpcli/http.go
```diff
@@ -79,8 +98,11 @@ func (c *Client) Do(m encoding.Marshaler) (encoding.Decoder, io.Closer, error) {
 	req.Header.Set("Content-Type", c.codec.MediaTypes[0])
 	req.Header.Set("Accept", c.codec.MediaTypes[1])
 
-	res, err := c.cli.Do(req)
+	res, err := c.do(req)
 	if err != nil {
+		if res != nil && res.Body != nil {
```
tsenart: I don't get this. When there is an error, res should be always nil. This invariant must be part of the API.

jdef:
The semantics of the RoundTripper API are different than that of
http.Client.Do, specifically w/ respect to error handling. I wanted to be
very clear w/ respect to such expectations and felt that using the
RoundTripper API would be confusing in this regard.

httpcli/http.go
jdef: There is nothing in the API docs for http.Client.Do that says that if err
!= nil then res will be nil.

https://golang.org/pkg/net/http/#Client.Do

Or, am I mis-reading?

jdef:
related w/ respect to headers and "Connection: Close" https://issues.apache.org/jira/browse/MESOS-3601

PR #210, opened 2015-11-13
next: document the plan going forward

jdef:


DESIGN.md
```diff
@@ -0,0 +1,27 @@
...
+### Target Audience
+
+- Developer A: high-level framework writer
+  - wants to provide barebones initial configuration
+  - wants to consume a simple event stream
+  - wants to inject events into a managed event loop (easy state machine semantics)
+
+```go
+c := client.New(connectionOptions..) // package scheduler/client
+eventCh, errCh := c.Register(registrationOptions..)
```
s-urbaniak: I am not sure of libraries (in this case `mesos-go`) returning channels. We should leave concurrency handling to the consumers of the library.

What I'd rather suggest is to leave mesos-go blocking as much as possible, like the `net/http` client package and simply have a blocking call like
```
event, err := c.Event()
```

s-urbaniak: Some more pointers in case we insist on using an event channel:

- https://github.com/mesosphere/mesos-dns/blob/master/detect/masters.go#L36
- https://www.quora.com/Are-all-Golang-libraries-non-blocking-and-async

tsenart: +1

The original idea was to have schedulers just decode events in a for loop. I realised I had some lingering code in my computer which I just pushed to https://github.com/mesos/mesos-go/pull/211

jdef: Yes, the example scheduler code that you posted fits very well with what I
had in mind for "Developer B".

However there's some common helpful things that mesos-go could provide as
well for Developer A -- but I don't want to necessarily force every
framework writer to use them:

- basic implementation of the event loop as part of a simple state machine,
that;
- manages auto-reconnection/re-registration to the latest mesos master
- sends appropriate events back to the scheduler in the event of a
disconnect during a pending call (e.g. taskLaunch)
- additional support for management of reserved resources (haven't played
with these, but from what I've seen there are some useful patterns to adopt
here)

Developer B can handle all of this themselves if they want to. But it would
be nice to eventually implement higher-level support things for Developer A
as well. We can certainly focus on the Developer B stuff first.

jdef: I'd agree that this is the way to go to support Developer B.

When building out higher level abstractions to support Developer A I think
chan's could be a great way to communicate events and state changes.

On Fri, Nov 13, 2015 at 3:24 AM, Sergiusz Urbaniak <notifications@github.com

jdef:
link to canonical design doc: https://github.com/apache/mesos/blob/master/docs/scheduler-http-api.md

PR #213, opened 2015-11-30
Fail back to hostname when hostname -f fails.

tnachen:
Hostname -f can fail when running in host container network, and there go bindings cannot work in that scenario.


jdef:
this is less reliable than specifying a hostname-override before go1.5, especially on windows: https://github.com/golang/go/issues/9982

any reason to not use a hostname-override instead?

jdef:
xref #214

jdef:
superseded by #218

PR #218, opened 2016-01-15
fix for: hostname resolution has binary deps on the host

### LINKED ISSUE #214: hostname resolution has binary deps on the host
the bindings may attempt to resolve the node's FQDN via `hostname -f` which likely isn't supported in all scenarios. also, `os.Hostname()` has known issues on windows before go1.5

we need a better strategy for falling back to something reliable for hostname resolution that works cross-platform.

jdef:
* old behavior: execute `hostname -f` and return the result, fatal on error
* new behavior: execute `hostname -f` and if that fails then call Go's `os.Hostname()`; fatal if nothing succeeds

### DIFF
--- a/mesosutil/node.go
+++ b/mesosutil/node.go
@@ -1,6 +1,7 @@
 package mesosutil
 
 import (
+	"os"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jdef:
@tnachen PTAL - I think this is what you had intended; this version should compile so that I can get this functionality merged

jdef:
travis failed due to a known test flake. this looks good to go. Tim?

tnachen:
LGTM!

PR #222, opened 2016-02-13
relocate mocks

jdef:
mocks that use stretchr/testify/mock end up pulling in global testing flags when imported mock structs are part of a non `_test` package. this PR relocates mocks to subpackages that should only be imported by `_test` package code.


jdef:
https://github.com/mesosphere/kubernetes-mesos/issues/785

PR #226, opened 2016-02-27
Added new API of NewScalarResourceWithRevocable.

jay-lau:


mesosutil/mesosprotoutil.go
```diff
@@ -36,6 +36,15 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
 	return result
 }
 
+func AddResourceRevocable(resource *mesos.Resource, principal string, role string) *mesos.Resource {
```
jdef: why are principal and role included as parameters here?

mesosutil/mesosprotoutil.go
```diff
@@ -36,6 +36,15 @@ func FilterResources(resources []*mesos.Resource, filter func(*mesos.Resource) b
...
+        resource.Revocable = &mesos.Resource_RevocableInfo{}
+        return resource
+}
+
+func NewScalarResourceWithRevocable(name string, value float64, principal string, role string) *mesos.Resource {
```
jdef: I'm not crazy about propagating this anti-pattern of func-name-for-every-resource-type-and-feature. This kind of API nonsense will disappear in mesos:next.

jdef:
thanks for the PR! I'll accept this once AddResourceRevocable is cleaned up

jay-lau:
@jdef It was updated in https://github.com/mesos/mesos-go/pull/227 , my mistake that I generate a new PR

PR #235, opened 2016-04-15
Mesos-Stream-Id header support

jdef:
* support for `Mesos-Stream-Id`
* minor httpcli API touchups
* example-scheduler flag/config cleanup
* support for Revive call

### DIFF
--- a/httpcli/http.go
+++ b/httpcli/http.go
@@ -78,21 +78,27 @@ func New(opts ...Opt) *Client {
 		do:     With(),
 		header: http.Header{},
 	}
-	return c.With(opts...)
+	c.With(opts...)
+	return c
 }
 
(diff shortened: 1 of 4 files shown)


httpcli/http.go
```diff
@@ -140,25 +155,89 @@ func (c *Client) Do(m encoding.Marshaler, opt ...RequestOpt) (encoding.Decoder,
 		events = c.codec.NewDecoder(recordio.NewFrameReader(res.Body))
 	case http.StatusAccepted:
 		// noop; no data to decode for these types of calls
+	case http.StatusTemporaryRedirect:
+		// TODO(jdef) refactor this
```
jdef: TODO

PR #268, opened 2016-11-12
Removed hardcoded type assertion in decoder.go

cinchurge:
framing.Decoder.Decode currently has a hardcoded type assertion for protocol buffers. This change will allow json unmarshalers as well.


jdef:
LGTM, thanks!

PR #269, opened 2016-11-16
Prevent panic on nil driver

relistan:
This shouldn't really happen in a real scenario, but the code probably shouldn't crash in this circumstance. If the driver is nil and `Start()` is still called, it will panic because no `nil` check was being done against the driver. This behavior happens e.g. in the example executor when not run from Mesos.

Steps to reproduce:

1. Compile example executor
2. Execute it
3. Panic


relistan:
Closing this because the application should be checking this. Looking at this further, it looks like the real bug is further upstream where 
```driver, err := executor.NewMesosExecutorDriver(dconfig)``` 
should not be returning a nil `driver` and nil `error`.

PR #278, opened 2017-02-13
next: Caller api for executors

jdef:
This brings the executor API into alignment w/ the `Caller` API that's available for schedulers


jdef:
superseded by #301

PR #284, opened 2017-04-13
Makefile and protobuf tweaks

jdef:
* fix broken `make protobufs`
* gogo customname some fields to be more idiomatic in their Golang form
* revisit gogo `nullable` for fields that reference `enum` types. document current thinking
* compare against latest Mesos 1.0.x tree for differences -- none found!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,14 @@
+mkfile_path	:= $(abspath $(lastword $(MAKEFILE_LIST)))
+current_dir	:= $(patsubst %/,%,$(dir $(mkfile_path)))
+
 MESOS_API_VERSION := v1
 API_PKG    := ./api/${MESOS_API_VERSION}/lib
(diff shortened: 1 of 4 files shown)


jdef:
fun failure w/ golang-tip:
```
13.83s$ make install
go install github.com/mesos/mesos-go/api/v1/cmd/example-executor github.com/mesos/mesos-go/api/v1/cmd/example-scheduler
# github.com/mesos/mesos-go/api/v1/lib/httpcli/httpsched
api/v1/lib/httpcli/httpsched/state.go:73:19: internal compiler error: cannot inline function containing closure: 
.   CLOSURE l(44) ff(1) tc(1) httpsched.maybeLogged.func1 FUNC-func(*http.Request) (*http.Response, error)
```

I'd really like to clean up that module anyway. Funny that golang has a problem with it

Issue #277, opened 2017-02-06
next: document or improve handling of master disconnections


ajwerner:
Right now the httpsched doesn't deal with disconnections from the master particularly well. Future attempts to re-subscribe on the same httpsched return errAlreadySubscribed. Currently my answer is to detect the error then mark the Context as Done and create a new httpsched. Is this the right approach? 

https://github.com/mesos/mesos-go/blob/next/httpcli/httpsched/state.go#L123

jdef:
That's probably the best way to handle it for now. Need to carefully consider how much additional intelligence should be built into the httpsched client vs. delegating to frameworks.

jdef:
fixed by #286

PR #286, opened 2017-04-16
scheduler improvements and API tweaks

jdef:
* httpsched transitions to disconnected upon errors generated by the Decoder
* demo scheduler properly handles Mesos master failover
* httpcli: deprecate ProtocolError and Err variables in favor of APIError
* minor doc tweaks


api/v1/lib/scheduler/calls/caller.go
```diff
@@ -85,6 +101,19 @@ func FrameworkCaller(frameworkID string) Decorator {
 	}
 }
 
+// DynamicFrameworkCaller generates and returns a Decorator that applies the given frameworkID to all calls (except SUBSCRIBE).
+func DynamicFrameworkCaller(frameworkID func() string) Decorator {
```
jdef: still dislike this name. maybe `SubscribedCaller` instead?

jdef: still dislike this name. maybe `SubscribedCaller` instead?

api/v1/cmd/example-scheduler/app/app.go
```diff
@@ -104,19 +121,21 @@ func buildEventHandler(state *internalState) events.Handler {
...
 				return nil
 			},
-			scheduler.Event_SUBSCRIBED: func(e *scheduler.Event) (err error) {
+			scheduler.Event_SUBSCRIBED: func(e *scheduler.Event) error {
 				log.Println("received a SUBSCRIBED event")
-				if state.frameworkID == "" {
-					state.frameworkID = e.GetSubscribed().GetFrameworkID().GetValue()
+				frameworkID := e.GetSubscribed().GetFrameworkID().GetValue()
+				if state.frameworkID == "" || state.frameworkID != frameworkID {
+					if state.frameworkID != "" && state.frameworkID != frameworkID && state.config.checkpoint {
```
jdef: not required to be nested, pull outside of containing `if`

api/v1/cmd/example-scheduler/app/config.go
```diff
@@ -84,6 +88,7 @@ func NewConfig() Config {
 		url:              env("MESOS_MASTER_HTTP", "http://:5050/api/v1/scheduler"),
 		codec:            codec{Codec: &encoding.ProtobufCodec},
 		timeout:          envDuration("MESOS_CONNECT_TIMEOUT", "20s"),
+		failoverTimeout:  envDuration("FRAMEWORK_FAILOVER_TIMEOUT", "1000h"),
```
jdef: SCHEDULER_FAILOVER_TIMEOUT

api/v1/lib/extras/scheduler/controller/controller.go
```diff
@@ -78,7 +76,9 @@ func (_ *controllerImpl) Run(config Config) (lastErr error) {
 	for !config.Context.Done() {
 		frameworkID := config.Context.FrameworkID()
 		if config.Framework.GetFailoverTimeout() > 0 && frameworkID != "" {
-			subscribe.Subscribe.FrameworkInfo.ID = &mesos.FrameworkID{Value: frameworkID}
+			frameworkProto := &mesos.FrameworkID{Value: frameworkID}
+			subscribe.Subscribe.FrameworkInfo.ID = frameworkProto
```
jdef: would be nice to consolidate these lines and say something like `subscribe.WithFrameworkID(frameworkID)`

api/v1/lib/httpcli/http.go
```diff
@@ -54,21 +57,45 @@ var (
...
+			Code:    code,
+			Message: err.Error(),
 		}
-		return err
 	})
 )
 
 // ProtocolError is a generic error type returned for expected status codes
 // received from Mesos.
+// Deprecated: no longer used in favor of APIError.
```
jdef: Might be better to just break the API now and remove this, otherwise downstream consumers will wonder why they're not seeing ProtocolError's anymore. Better to break the build and force change.

Maybe tackle this in a follow-up PR? This one is getting a bit long already.

jdef:
upon the scheduler receiving an ERROR event, it should probably re-try subscription. unless Mesos terminates the current subscription connection after sending an ERROR (which is not in the contract) then we're still in trouble because the `httpsched` state doesn't know anything about these ERROR messages so it will stay "connected" (and reject SUBSCRIBE attempts)

jdef:
This PR is massive enough and I'm mostly happy (enough) with it. Will keep it open for a few days for comments and merge if there are no complaints.

jdef:
ideally there would be tests shipping with these changes

jdef:
fixes #277

jdef:
filed ticket to track tech debt for better unit tests in `httpcli` and `httpsched`: #291

PR #283, opened 2017-03-14
Switch to glide in favor of submodules for dependency management

cinchurge:
Switch to glide for managing vendor libraries instead of using submodules which are much more painful to manage.

PR #287, opened 2017-04-18
replace submodule vendoring with govendor

jdef:


jdef:
leaving open for a few days to gather feedback before mergine

pires:
LGTM.

Related, would you be open to adopt [`glide`](https://glide.sh/) for dependency management?

pires:
Nevermind, just noticed #283.

PR #292, opened 2017-04-24
API package godocs that summarize and favor the v1 API over v0

jdef:


pires:
Just my two cents, but I don't like `api/v1/lib` subpackage. I'd move [examples as part of the `api/v1` package](https://blog.golang.org/examples) where possible. In case an example doesn't make sense or doesn't work following that method, move it to `/examples/v1`.

jdef:
The examples are pretty complex. `examples/{v0,v1}` is something I could get behind. Though I would sort of feel bad reorganizing the repo twice in such a short time period.

jdef:
proposal for (yet again) restructuring the repo: #294

pires:
That's what we, software engineers, do all the time: move things around and call it fancy names like refactor, redesign, restructure. Do it 👍

jdef:
committing this change to offer guidance until restructuring takes place

PR #297, opened 2017-04-28
executor backoff typo fix (3/4 instead of 4/3)

### LINKED ISSUE #295: executor backoff calculation should use a factor of 3/4 instead of 4/3
as per convo https://github.com/mesos/mesos-go/commit/94f43cc3ef0c498af819207c3e03ceaddc5b73ca#commitcomment-21894591

kwiatkk1:
Fixes #295


jdef:
thanks for the PR!

PR #298, opened 2017-04-28
httpsched: added unit test for disconnect decoder

jdef:


jdef:
awesome: added a unit test and coverage drops :(

PR #303, opened 2017-06-03
apierrors: make response message more descriptive

hotienvu:
Something like *malformed request: Failed to validate scheduler::Call: Expecting 'framework_id' to be present* is more helpful than just "malformed request"


jdef:
thanks for the PR! I'd prefer a bit cleaner output if `Details` isn't set. How about changing to something like:
```go
if err.Details != "" {
  return err.Message + ": " + err.Details
}
return err.Message
```

hotienvu:
I have fixed and added tests but coverage still drops. It looks like the coverage settings need to be updated as well?

jdef:
don't worry about coverage, i've got a fix for that coming

PR #304, opened 2017-06-24
protos: upgrade to latest 1.2.x protobuf API

jdef:
- update mesos protos to the latest 1.2.x v1 API
- scheduler and executor API updates to correspond to proto API changes
- update examples to comply with mesos 1.2.x API
- patched a bug in the `msh` example

### DIFF
--- a/api/v1/cmd/example-scheduler/app/state.go
+++ b/api/v1/cmd/example-scheduler/app/state.go
@@ -29,6 +29,7 @@ func prepareExecutorInfo(
 	if execImage != "" {
 		// Create mesos custom executor
 		return &mesos.ExecutorInfo{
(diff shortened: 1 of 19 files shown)


jdef:
MESOS change for Revive/Suppress messaging: https://issues.apache.org/jira/browse/MESOS-6638

`SUPPRESS` and `REVIVE` were already members of `Call.Type` prior to 1.2; now that they have corresponding message object types, it's unclear if the message objects are actually required if there's no `role` to set. The docs are unclear on this

jdef:
As per the above, message objects are NOT required for `SUPPRESS` and `REVIVE` in the 1.2 API (and a quick scan of the 1.3.x impl shows the same): https://github.com/apache/mesos/blob/1.2.x/src/master/validation.cpp#L382

PR #305, opened 2017-06-25
Upgrade golang versioin to 1.8.1

keloyang:
With golang1.6.1, when 'make docker', we will encounter the following issue,
```
api/v1/cmd/example-executor/main.go:4:2: cannot find package "context" in any of:
        /go/src/github.com/mesos/mesos-go/api/v1/cmd/vendor/context (vendor tree)
        /go/src/github.com/mesos/mesos-go/api/v1/vendor/context
        /usr/local/go/src/context (from $GOROOT)
        /go/src/context (from $GOPATH)
Makefile:108: recipe for target 'docker' failed
```
This is because golang has no package "context", so upgrade golang to 1.8.1 .

Signed-off-by: Shukui Yang <yangshukui@huawei.com>


jdef:
thanks for the PR!

PR #314, opened 2017-09-05
Add support for multiple disks resources

everpcpc:
Disk source should also be checked. Otherwise it will cause a mistaken resource comparison when using multiple disks.

### DIFF
--- a/api/v1/lib/resources.go
+++ b/api/v1/lib/resources.go
@@ -229,21 +229,21 @@ func (resources Resources) String() string {
 			if s := d.GetSource(); s != nil {
 				switch s.GetType() {
 				case PATH:
-					buf.WriteString("(PATH:")
+					buf.WriteString("PATH:")
 					if p := s.GetPath(); p != nil {
 						buf.WriteString(p.GetRoot())
 					}
 				case MOUNT:
-					buf.WriteString("(MOUNT:")
+					buf.WriteString("MOUNT:")
 					if m := s.GetMount(); m != nil {
 						buf.WriteString(m.GetRoot())
 					}
-				default:
-					buf.WriteString("(ROOT:")
 				}
-				buf.WriteString(")")
 			}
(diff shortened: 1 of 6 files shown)


jdef:
Thanks for the PR! Would you mind fleshing out `TestResources_PersistentVolumes` or else writing a new unit test to exercise this new code?

everpcpc:
Hi, I've put the tests in `TestResources_Equivalent`.

api/v1/lib/resources.go
```diff
@@ -226,6 +226,23 @@ func (resources Resources) String() string {
...
+					if p := s.GetPath(); p != nil {
+						buf.WriteString(p.GetRoot())
+					}
+				case MOUNT:
+					buf.WriteString("(MOUNT:")
+					if m := s.GetMount(); m != nil {
+						buf.WriteString(m.GetRoot())
+					}
+				default:
+					buf.WriteString("(ROOT:")
```
jdef: can we get rid of `ROOT:` output here? I'd like for the output to be mostly consistent with https://github.com/apache/mesos/blob/1.2.x/src/common/resources.cpp#L1922

api/v1/lib/resourcetest/resourcetest.go
```diff
@@ -72,7 +72,7 @@ func Reservation(ri *mesos.Resource_ReservationInfo) Opt {
 	}
 }
 
-func Disk(persistenceID, containerPath string) Opt {
+func Disk(persistenceID, containerPath, source string, sourceType mesos.Resource_DiskInfo_Source_Type) Opt {
```
jdef: nit: would rather have left the params for `Disk` alone and created a new `DiskWithSource` func w/ the extended params. Callers of `Disk(x, y, "", 0)` shouldn't actually need to bother with the additional params.

jdef:
thanks, lgtm

PR #325, opened 2017-09-18
executor: Sender API support, updated example-executor

### LINKED ISSUE #317: add api/v1/lib/executor/calls Sender API
xref #312 (ala api/v1/lib/agent/calls.Sender)

jdef:
fixes #317


api/v1/cmd/example-executor/main.go
```diff
@@ -104,43 +111,35 @@ func run(cfg config.Config) {
...
-		if n := len(state.unackedTasks); n > 0 {
-			unackedTasks := make([]mesos.TaskInfo, 0, n)
-			for k := range state.unackedTasks {
-				unackedTasks = append(unackedTasks, state.unackedTasks[k])
-			}
-			call.Subscribe.UnacknowledgedTasks = unackedTasks
-		} else {
-			call.Subscribe.UnacknowledgedTasks = nil
+// unacknowledgedTasks generates the value of the UnacknowledgedTasks field of a Subscribe call.
+func unacknowledgedTasks(state *internalState) (result []mesos.TaskInfo) {
```
janisz: I think `unacknowledgedTasks` should be a function on `*internalState` but it's a minor issue.

jdef: I could go either way. Maybe tackle in a follow-up PR

PR #330, opened 2017-10-23
framing/decoding related fixes

### LINKED ISSUE #329: recordio: clearly distinguish between EOF and EOF-no-data cases
Currently there's no way for a client of a decoder that's working against a recordio stream to distinguish between the (frame,EOF) case and (no-frame,EOF) case. As a result, an Unmarshaler may be invoked with a zero-length frame, which doesn't play well (at least, for the built-in JSON unmarshaler).

jdef:
fixes #329

### DIFF
--- a/api/v1/lib/encoding/framing/decoder_test.go
+++ b/api/v1/lib/encoding/framing/decoder_test.go
@@ -1,4 +1,4 @@
-package framing
+package framing_test
 
 import (
 	"errors"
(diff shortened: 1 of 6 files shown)


api/v1/lib/encoding/framing/decoder_test.go
```diff
@@ -28,40 +30,93 @@ func TestNewDecoder(t *testing.T) {
...
+					f = append(f, buf)
+				}
+				if err != tc.wantsErr {
+					t.Errorf("expected error %q instead of %q", tc.wantsErr, err)
+				}
+			}
+			if !reflect.DeepEqual(f, tc.wants) {
+				t.Errorf("expected %#v instead of %#v", tc.wants, f)
+			}
+			if (tc.wants == nil) != (len(f) == 0) {
```
jdef: this `if` block looks redundant: the DeepEqual check above already does this

PR #332, opened 2017-11-02
Prohibit http.Client from following internal redirects

rboyer:
By prohibiting internal redirects inside of http.Client, we can let the framework code actually witness the 307 non-master redirects and update endpoint logic accordingly.


jdef:
thanks for the PR! LGTM

Issue #333, opened 2017-11-02
httpsched contains debug print statements


rboyer:
The `api/v1/lib/httpcli/httpsched/httpsched.go` file contains several bare `log.Printf` statements that look to be stray debugging statements.

After https://github.com/mesos/mesos-go/pull/332 is fixed, these will be surfaced because the redirect handling code was functionally dead before, so these were not peppering stderr.

PR #334, opened 2017-11-02
mute httpsched debug logging

rboyer:
Repurpose the existing `debug` package-local flag to suppress some debug logs that are currently not suppressed.

Fixes https://github.com/mesos/mesos-go/issues/333


jdef:
thanks for the PR! LGTM

Issue #255, opened 2016-07-19
next: can not resubscribe


athlum:
Hi jdef,
https://github.com/mesos/mesos-go/blob/next/httpcli/httpsched/state.go#L123
I find there is no way to change `connectedFn` to `disconnectedFn `. So the framework couldn't call subscribe (with the frameworkId) again when connection broken.
And `stateFn` switches on every call only. I want to know how would you deal with `stateFn` change.
Thanks.

PR #337, opened 2017-11-08
feature flag opt-in for re-subscribe

rboyer:
I noticed that the v1 scheduler workflow isn't re-subscribe friendly, which makes implementing proper heartbeat handling in a framework tricky.

This PR adds the bare minimum hooks necessary to test out re-SUBSCRIBE on an opt-in basis. It "should" be safe to do without the feature-flag, but the comments by jdef in the source imply otherwise.

Doing this without the opt-in does have repercussions on anyone who is currently invoking a SUBSCRIBE call. To avoid issues with receiving one last `scheduler.Event_ERROR` message on the old subscription socket, you have to do some additional state tracking in framework code to correctly handle it.  As I'm not using the helpers in the `extras` directory, I'm not sure the right way to handle the workflow there so I didn't make any changes to try and use the re-subscribe changes there.

### DIFF
--- a/api/v1/lib/httpcli/httpsched/httpsched.go
+++ b/api/v1/lib/httpcli/httpsched/httpsched.go
@@ -37,7 +37,8 @@ type (
 
 	client struct {
 		*httpcli.Client
-		redirect RedirectSettings
+		redirect       RedirectSettings
+		allowReconnect bool // feature flag
 	}
 
 	// Caller is the public interface a framework scheduler's should consume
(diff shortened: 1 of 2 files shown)


jdef:
Thanks for this PR! The scheduler not receiving heartbeat signals from the master are a good example of a condition not picked up by the default error detection in the httpsched impl.

jdef:
#255

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -96,6 +97,14 @@ func MaxRedirects(mr int) Option {
 	}
 }
 
+func AllowReconnection(v bool) Option {
```
jdef: mind writing a minimal comment for this func that explains how it effectively alters the behavior of the client?

jdef: i'm going to add comments in a subsequent PR

Issue #340, opened 2018-01-16
v1: gogoproto annotations prevent valid reservation request generation


ajwerner:
In mesos 1.4 the agent API enforces the comment on ReservationInfo that
```
// NOTE: This field must not be set for `Resource.reservation`.
```
https://github.com/mesos/mesos-go/blob/master/api/v1/lib/mesos.proto#L1253
but because of the 
```
optional Type type = 4 [(gogoproto.nullable) = false];
```
it is impossible to omit `type`. 

so we end up getting back a 400 response from the agent:
```
Header:  map[Date:[Wed, 10 Jan 2018 16:52:46 GMT] Content-Length:[117] Content-Type:[text/plain; charset=utf-8]]
Status:  400 Bad Request
Body:  Resource 'mem:30000' is invalid: 'Resource.ReservationInfo.type' must not be set for the 'Resource.reservation' field
```
We've verified that setting nullable to true, regenerating the protos, and setting type to nil resolves the issues.

ajwerner:
this appears to be addressed by [https://github.com/mesos/mesos-go/pull/335](this pull request)

ajwerner:
#335

ajwerner:
What's the ETA #335? If it's more than a couple of weeks, would you be willing to accept a pull request with the single line change?

jdef:
Thanks for this, yes willing to accept a 1 line change. I'm holding off on
1.5.x protos until Mesos 1.5 actually lands.

jdef:
resolved by #341

PR #341, opened 2018-01-19
lib: allow ReservationInfo.Type to be nil

ajwerner:
This is to address incompatibility with Mesos 1.4 agent reservations as described in #340


jdef:
thanks for the PR! it looks like the gogo protobuf generation tools are out-of-sync w/ the version of the gogo tooling that's been vendored by the project. please see the travis config for how to install the right version of the tooling and then please re-generate the protos.

https://github.com/mesos/mesos-go/blob/c488712c2b5a908fc9534560386fc572c3d126bf/.travis.yml#L19

ajwerner:
Should be good to go

jdef:
thanks!

Issue #339, opened 2017-11-10
v1: add support for multiple endpoints


rboyer:
If you run a replicated mesos-master setup (say N=3) then the current state of how you inform the HTTP API client of the masters is:

1. pick one, and use `httpcli.Endpoint` to save it
2. hope it never goes down
3. let the 307 redirection code ask any master for the leading master and update your favorite

This works ok enough until the selection from (1) is not responsive. Then you have to cobble together some sort of much higher level round-robin over a list of mesos-master addresses.

It would be cool if the `httpcli.Client` could take one of the following:
```
// NextEndpointFunc returns a new endpoint to try.  
type NextEndpointFunc func() string

// example of NextEndpointFunc
func simpleRoundRobin(urls []string) httpcli.NextEndpointFunc {
       var pos int
       return func() string {
               if pos >= len(urls) {
                       pos = 0
               }
               s := urls[pos]
               pos++
               return s
       }
}
```
```
// EndpointProviderFunc could hook into a service-discovery option or flat config file
type EndpointProviderFunc func() []string

// example of EndpointProviderFunc from a flat file:
func staticHosts(urls []string) httpcli.EndpointProviderFunc {
    return urls
}
```

And then have it rotate through the options down wherever `c.url` is referenced when there's an error indicative of a dead or bad master.

rboyer:
This would relate to https://github.com/mesos/mesos-go/issues/338

jdef:
I like the proposed interface. Which specific failures would trigger calls to such an interface? "no such host"? "connection timeout"? something else?

mlowicki:
https://github.com/mesos/mesos-go/pull/352 is marked as merged. Does it mean that this ticket has been solved?

jdef:
Yes I believe this is resolved. Closing out. Please re-open if you feel that something here has been left unaddressed.

PR #352, opened 2018-06-20
support multiple master candidates

huahuiyang:
In this pr, if the mesos framework specify a candidate selector(could be dynamic func), the framework will gain the ability to failover across multiple mesos masters. otherwise, it will only rely on the url as an initial configuration.

### DIFF
--- a/api/v1/lib/httpcli/httpsched/httpsched.go
+++ b/api/v1/lib/httpcli/httpsched/httpsched.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/mesos/mesos-go/api/v1/lib"
@@ -173,15 +174,27 @@ func (cli *client) httpDo(ctx context.Context, m encoding.Marshaler, opt ...http
 	opt = append(opt, httpcli.Context(ctx))
 	for attempt := 0; ; attempt++ {
 		resp, err = cli.Client.Do(m, opt...)
-		redirectErr, ok := err.(*mesosRedirectionError)
-		if !ok {
-			return resp, err
+		if err == nil {
+			return
 		}
+		redirectErr, ok := err.(*mesosRedirectionError)
+
 		if attempt < cli.redirect.MaxAttempts {
-			if debug {
-				log.Println("redirecting to " + redirectErr.newURL)
+			var newURL string
+			if !ok {
+				candidates := strings.Split(cli.Candidates(), ",")
+				if len(candidates) <= 0 {
+					log.Printf("not found candidate urls, return directly")
+					return
+				}
+				newURL = candidates[attempt%len(candidates)]
+			} else {
+				newURL = redirectErr.newURL
 			}
-			cli.With(httpcli.Endpoint(redirectErr.newURL))
+			if debug {
+				log.Printf("redirecting to %v", newURL)
+			}
+			cli.With(httpcli.Endpoint(newURL))
 			select {
 			case <-getBackoff():
 			case <-ctx.Done():
--- a/api/v1/lib/httpcli/httpsched/httpsched.go
+++ b/api/v1/lib/httpcli/httpsched/httpsched.go
@@ -5,7 +5,6 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-	"strings"
 	"time"
 
 	"github.com/mesos/mesos-go/api/v1/lib"
@@ -41,6 +40,7 @@ type (
 		redirect       RedirectSettings
 		allowReconnect bool // feature flag
 		listener       func(Notification)
+		candidates     []string
 	}
 
 	// Caller is the public interface a framework scheduler's should consume
@@ -135,6 +135,14 @@ func AllowReconnection(v bool) Option {
 	}
 }
 
+func MasterCandidates(masters []string) Option {
+	return func(c *client) Option {
+		old := c.candidates
+		c.candidates = masters
+		return MasterCandidates(old)
+	}
+}
+
 // NewCaller returns a scheduler API Client in the form of a Caller. Concurrent invocations
 // of Call upon the returned caller are safely executed in a serial fashion. It is expected that
 // there are no other users of the given Client since its state may be modified by this impl.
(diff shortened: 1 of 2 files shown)


hello-yhh:
this is to support multiple master api urls when start the framework. 
```bash
cli := httpcli.New(
		httpcli.Candidates(cfg.Masters),
		httpcli.Codec(cfg.Codec),
		httpcli.Do(httpcli.With(
...
```

huahuiyang:
this is to support multiple master api urls when start the framework, support multiple mesos api urls failover in attempts
```bash
cli := httpcli.New(
		httpcli.Candidates(cfg.Masters),
		httpcli.Codec(cfg.Codec),
...
```

huahuiyang:
related https://github.com/mesos/mesos-go/issues/339

huahuiyang:
@jdef  would you like to take a look when you get a chance?

huahuiyang:
@tsenart  @vladimirvivien @jdef any ideas? does this repo still in maintaining? seems it has been a long time after this pr was created.

jdef:
Thanks for the PR! This repo *is* maintained, and I'm basically the only maintainer at this point .. so sometimes it takes me while to get around to reviewing PRs. Thanks for being patient.

It looks like the use case you're trying to address is this:

1. A cluster has multiple master candidates
2. The default HTTP client is expected to always hit some "primary" master candidate (the first one in the list); if this master is down then calls will fail.
3. The scheduler HTTP client will round-robin across all candidates if an API call results in a non-redirection error; for redirects, continue to use the candidate suggested by the result returned by Mesos.

Does that sound about right?

These candidates are more like "bootstrap" endpoints, right? Because in a cloudy environment where servers come and go, it's possible that the initial nodes might cycle out (unless you were using floating IPs that remaining the same across recycled instances).

I'm interested in the specifics if your use case. Please elaborate in the PR description.

huahuiyang:
@jdef  thanks for seeing this pr, and the items you listed in the comment are pretty much correct! 
setup a load balancer in front of mesos masters, providing a non-change endpoint is a solution, but having some drawbacks:
1. a load balancer is a must have in that scenario (dns based service discovery takes time to sync, so we assume people using nginx/haproxy/lvs  etc. as a load balancer)
2. setup a load balancer to proxy active master, introduce a hop which might be better to mitigate if we care performance a lot. 
in our case, we choose to use mesos masters raw ip/dns directly (akka. candidates in this pr), we noticed that mesos master addresses could be changed totally, but in a high availability, fault tolerant mesos scheduler, the active/standby schedulers could be easily rolling restart to load the latest mesos master candidates in case.

jdef:
OK. Let me then suggest the following changes to this PR:

1. don't change the `httpcli` package at all since what you really want are changes to the `httpsched` client; create an `Option` func in `httpsched` handles candidate selection
2. loosen up the candidate specification: instead of some encoded string, maybe a candiate selector option is some kind of a `func() string`. then you could define a "static candidate selector" that cycles over some fixed slice of candidate strings. but a user could invent a new implementation that was more dynamic and not limited to an initial, fixed set of strings. let me know if you need/want me to elaborate on this idea more.

jdef:
e.g.
```go
type CandidateSelector func() string

func FixedCandidateSelector(s []string) CandidateSelector { ... } // cycles over s
```

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -173,15 +184,31 @@ func (cli *client) httpDo(ctx context.Context, m encoding.Marshaler, opt ...http
...
+		if err == nil {
+			return
 		}
+		redirectErr, ok := err.(*mesosRedirectionError)
+
 		if attempt < cli.redirect.MaxAttempts {
+			var candidate string
+			if !ok {
+				if cli.candidateSelector == nil {
+					log.Printf("not found candidate selector, using url when initilize framework")
```
jdef: please wrap calls to `log.XXX` with an `if debug` check

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -173,15 +184,31 @@ func (cli *client) httpDo(ctx context.Context, m encoding.Marshaler, opt ...http
...
+					candidate = cli.Endpoint()
+				} else {
+					candidate = cli.candidateSelector()
+				}
+				if candidate == "" {
+					log.Printf("not found candidate url, return directly")
```
jdef: ditto

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -173,15 +184,31 @@ func (cli *client) httpDo(ctx context.Context, m encoding.Marshaler, opt ...http
...
+					return
+				}
+			} else {
+				candidate = redirectErr.newURL
```
jdef: maybe make this the default value of `candidate` when it's initially declared and get rid of this `else` statement

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -74,6 +75,8 @@ type (
 		requestOpts    []httpcli.RequestOpt // requestOpts are temporary per-request options
 		opt            httpcli.Opt          // opt is a temporary client option
 	}
+
+	CandidateSelector func() string
```
jdef: yes, please add at least a one-line doc string for this. something like:
```go
// CandidateSelector returns the next endpoint to try if there are errors reaching the mesos master, or else an empty string if there are no such candidates.
```

api/v1/lib/httpcli/httpsched/httpsched.go
```diff
@@ -134,6 +137,14 @@ func AllowReconnection(v bool) Option {
 	}
 }
 
+func MasterCandidates(cs CandidateSelector) Option {
```
jdef: nit: maybe rename to `EndpointCandidates` for clarity (since it's actually selecting the next endpoint for the http client)

api/v1/lib/httpcli/httpsched/httpsched.go
huahuiyang: get newURL from redirectErr object is only safe when the ok == null. (unless there will be some nil point error)
this else block seems is necessary at here

huahuiyang:
great, after upstream this change, our prod system do not need to maintain a forked mesos-go repo for multiple candidates purpose in our organization. i will think of the unit test part in another pr.

jdef:
@huahuiyang please rebase so that I can merge this

huahuiyang:
@jdef rebased.

huahuiyang:
ping @jdef

huahuiyang:
@jdef  any chance for you to take a look and get this pr merged? thanks

jdef:
LGTM, thanks

PR #353, opened 2018-07-18
proto: bump to 1.7.x 8088142

chhsia0:
Updated Mesos protos to adapt the new `CREATE_DISK`/`DESTROY_DISK`. Also added a new master call for `GET_OPERATIONS`, and new agent calls for `GET_OPERATIONS`, `GROW_VOLUME` and `SHRINK_VOLUME`. For the latter two, only agent-local volumes are supported.


jdef:
... otherwise the changes look great!

api/v1/lib/mesos.proto
```diff
@@ -2058,10 +2051,8 @@ message Offer {
     optional Destroy destroy = 6;
     optional GrowVolume grow_volume = 13; // EXPERIMENTAL.
     optional ShrinkVolume shrink_volume = 14; // EXPERIMENTAL.
-    optional CreateVolume create_volume = 8;
```
jdef: why weren't these removed tag numbers `reserved` upstream to avoid possible future conflicts?

jdef: https://developers.google.com/protocol-buffers/docs/proto#reserved

jdef:
Please rebase this now that the 1.6.x protos have landed on master, thanks!

PR #354, opened 2018-07-18
proto: bump to 1.6.x 2f6f381

chhsia0:
Updated Mesos protos to 1.6.x, which adds a new master call for GET_OPERATIONS, and new agent calls for GET_OPERATIONS, GROW_VOLUME and SHRINK_VOLUME. For the latter two, only agent-local volumes are supported.


api/v1/lib/scheduler/scheduler.proto
```diff
@@ -239,7 +239,8 @@ message Response {
     repeated OperationStatus operation_statuses = 1 [(gogoproto.nullable) = false];
   }
 
-  optional ReconcileOperations reconcile_operations = 1;
+  optional Type type = 1 [(gogoproto.nullable) = false];
```
jdef: this tag change is unfortunate but aligns perfectly w/ upstream changes

api/v1/lib/master/calls/calls.go
```diff
@@ -221,6 +228,32 @@ func DestroyVolumes(a mesos.AgentID, v ...mesos.Resource) *master.Call {
 	}
 }
 
+// GrowLocalVolume grows a persistent volume on an agent's ROOT or PATH disks. The request is forwarded asynchronously
+// to the Mesos agent where the persistent volume is located.
+func GrowLocalVolume(a mesos.AgentID, volume mesos.Resource, addition mesos.Resource) *master.Call {
```
jdef: nit: it's more idiomatic to say `volume, addition mesos.Resource` (since they've of the same type)

jdef: also: mesos.AgentID is optional in the protos, so it should be a pointer here

api/v1/lib/master/calls/calls.go
```diff
@@ -221,6 +228,32 @@ func DestroyVolumes(a mesos.AgentID, v ...mesos.Resource) *master.Call {
...
+			AgentID:  &a,
+			Volume:   volume,
+			Addition: addition,
+		},
+	}
+}
+
+// ShrinkLocalVolume shrinks a persistent volume on an agent's ROOT or PATH disks. The request is forwarded
+// asynchronously to the Mesos agent where the persistent volume is located.
+func ShrinkLocalVolume(a mesos.AgentID, volume mesos.Resource, subtract mesos.Value_Scalar) *master.Call {
```
jdef: mesos.AgentID is optional in protos, should be a pointer parameter

jdef:
Thanks, left some feedback. Surprised that schedulerpb_test.go didn't have any changes

api/v1/lib/scheduler/scheduler.proto
chhsia0: I checked with Gaston and confirmed that this was not used anywhere when introduced, so it should be a "safe" change.

api/v1/lib/master/calls/calls.go
chhsia0: I'll make it a pointer. Then should I just rename it to `GrowVolume`/`ShrinkVolume`? Currently it supports only local volumes but me may extend it to support external volumes in the future.

jdef:
I'm fine w/ GrowVolume/ShrinkVolume

chhsia0:
@jdef Done. PTAL.

PR #355, opened 2018-07-19
Add go get to make install to fix a missing dependency

philipnrmn:
Without the `go get ./...` step, a newly configured environment does not
automatically import the ffjson and uuid libraries, causing `make install` to
fail due to missing dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ all: test
 .PHONY: install
 install: INSTALL_FLAGS +=
 install:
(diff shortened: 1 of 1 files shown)


jdef:
Thanks for the PR! It sounds like the dependencies are out of date. `make sync` should get everything needed to build the project. Instead of executing go-get, I think we should fix the dependencies so that `make sync` works as expected.

PR #362, opened 2018-09-28
scheduler: Jitter and RefuseDuration

jdef:


jdef:
Still haven't had a need for this, closing out. Will re-open later if needed.

PR #365, opened 2018-12-07
httpcli: Content-Length as obj type indicator

jdef:
For higher level clients using ResponseClassAuto (e.g. scheduler
callers) infer the response object type based on the Content-Length
header in the HTTP response. Streaming responses using recordio will not
have sent a Content-Length header, whereas "singleton" JSON object
responses will have one.


jdef:
context: the scheduler v1 API recently added the RECONCILE_OPERATIONS call, which returns a synchronous response (HTTP 200) w/ a JSON content type. The only other call prior to this that returned 200 and advertised a JSON content type was SUBSCRIBE (which actually returns recordio content).

PR #372, opened 2019-06-13
Add heartbeat helper method

bamarni:
This adds a helper method for sending heartbeat control messages with the `ATTACH_CONTAINER_INPUT` call.

### DIFF
--- a/api/v1/lib/agent/calls/calls.go
+++ b/api/v1/lib/agent/calls/calls.go
@@ -220,6 +220,22 @@ func AttachContainerInputTTY(t *mesos.TTYInfo) *agent.Call {
 	}
 }
 
+func AttachContainerInputHeartbeat(heartbeat *agent.ProcessIO_Control_Heartbeat) *agent.Call {
+	return &agent.Call{
+		Type: agent.Call_ATTACH_CONTAINER_INPUT,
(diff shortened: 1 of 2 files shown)


api/v1/lib/agent/calls/calls.go
```diff
@@ -220,6 +220,22 @@ func AttachContainerInputTTY(t *mesos.TTYInfo) *agent.Call {
 	}
 }
 
+func AttachContainerInputHeartbeat(heartbeat *agent.ProcessIO_Control_Heartbeat) *agent.Call {
```
jdef: Thanks for the PR. Mind updating call_test.go too?

api/v1/lib/agent/calls/calls.go
```diff
@@ -3,7 +3,7 @@ package calls
 import (
 	"time"
 
-	"github.com/mesos/mesos-go/api/v1/lib"
+	mesos "github.com/mesos/mesos-go/api/v1/lib"
```
jdef: nit: we don't do this anywhere else, why here?

bamarni: No real reason, I use vscode and it keeps on updating it whenever I save the file, I'll revert.

PR #378, opened 2019-08-01
[WIP] Go module support

nfnt:
Use Go modules, but keep a `vendor` directory for older versions of Go.


.travis.yml
```diff
@@ -1,23 +1,19 @@
...
+  - 1.12.x
 before_install:
    #these two lines help users who fork mesos-go. It's a noop when running from the mesos organization
   - RepoName=`basename $PWD`; SrcDir=`dirname $PWD`; DestDir="`dirname $SrcDir`/mesos"
   - if [[ "$SrcDir" != "$DestDir" ]]; then mv "$SrcDir" "$DestDir"; cd ../../mesos/$RepoName; export TRAVIS_BUILD_DIR=`dirname $TRAVIS_BUILD_DIR`/$RepoName; fi
   - go get github.com/mattn/goveralls
-  - go get github.com/kardianos/govendor
-  - make sync
-  - api/v1/vendor/github.com/gogo/protobuf/install-protobuf.sh
+  - vendor/github.com/gogo/protobuf/install-protobuf.sh
```
nfnt: TODO: This might not work anymore, we could, e.g., `curl` this instead.

jdef:
Thanks for investigating this. I briefly skimmed the changed files and noticed that the nested, vendored deps were converted to a top level deps, of a top level mod. This isn't consistent with keeping v0 api deps separate from v1 api deps, or separating the deps of the lib from the cmd examples. 

This repo is ripe for an overhaul, part of that should include support for gomods. Let's discuss a more comprehensive plan before moving forward with this PR.

nfnt:
Yes, I did convert them temporarily. The main work is making sure that Travis is still working as expected and steps that were depending on `govendor` don't break/ are changed accordingly. Once that's done, I wanted to separate the deps again.

PR #379, opened 2019-08-06
Add a Code() method to apierrors.Error

bamarni:
Being able to get the status code behind a Mesos error is necessary in order to fix:

https://jira.mesosphere.com/browse/DCOS_OSS-5432


bamarni:
Hum, not sure why it decreased, the method is covered in the test.

jdef:
The Code type already has a Matches method, not sure why this addition is
needed.

bamarni:
@jdef : indeed I had missed this method, seems like I can create a 500 Code and invoke Match(err) as an alternative. Thanks!

